PR #25, opened 2025-06-11
Add support for other backends, such as OpenRouter and Ollama

maxer137:
This PR introduces support for selecting and using different OpenAI-compatible backends. Users can now select from official OpenAI API, OpenRouter, or a local Ollama instance for their LLM calls.

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -444,20 +444,29 @@ def get_user_selections():
     )
     selected_research_depth = select_research_depth()
 
-    # Step 5: Thinking agents
+    # Step 5: OpenAI backend
     console.print(
         create_question_box(
-            "Step 5: Thinking Agents", "Select your thinking agents for analysis"
+            "Step 5: OpenAI backend", "Select which service to talk to"
         )
     )
-    selected_shallow_thinker = select_shallow_thinking_agent()
-    selected_deep_thinker = select_deep_thinking_agent()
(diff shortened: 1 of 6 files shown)


bugsnatch:
nice work! I was working on the same however I was planning to switch to langchain-ollama instead of openai library. How did you tackle the issue of web search? I see on this PR execution stops there due mismatch endpoint with ollama  

tried with qwen3

Yijia-Xiao:
Hi @maxer137,

Thank you for the PR! I will review and merge soon.

Yijia-Xiao:
> nice work! I was working on the same however I was planning to switch to langchain-ollama instead of openai library. How did you tackle the issue of web search? I see on this PR execution stops there due mismatch endpoint with ollama
> 
> tried with qwen3

Hi @bugsnatch,

Thank you for the contribution!

About the search part, in our internal version, we used some news/social media sources.

However, since those APIs require registration, and we hope to make it easier for users to set up, we did not include those tools in the release.

- One solution is to manually add `search tools` to models like `Qwen`.

- Another solution is to directly use news/social media tools as information sources. @EdwardoSunny can provide the list of tools that can be plugged in.

saksham0161:
@Yijia-Xiao may be we can build search tools and in readme add them as a optin feature? (If onboarding becomes a pain)

Issue #27, opened 2025-06-11
Bug in prompting


saksham0161:
I selected for TEAM stock analysis however my openai logs show that prompting has been closely related to TSLA

![Image](https://github.com/user-attachments/assets/9470fbe3-8b29-4c58-9192-f251f5396d0f)

![Image](https://github.com/user-attachments/assets/abacbb08-b932-45a0-b789-ab2c26f4d7a5)

saksham0161:
`│                                                                              │
│  Step 1: Ticker Symbol                                                       │
│  Enter the ticker symbol to analyze                                          │
│  Default: SPY                                                                │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
 [SPY]: TEAM
╭──────────────────────────────────────────────────────────────────────────────╮
│                                                                              │
│  Step 2: Analysis Date                                                       │
│  Enter the analysis date (YYYY-MM-DD)                                        │
│  Default: 2025-06-11                                                         │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
 [2025-06-11]:
╭──────────────────────────────────────────────────────────────────────────────╮
│                                                                              │
│  Step 3: Analysts Team                                                       │
│  Select your LLM analyst agents for the analysis                             │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
? Select Your [Analysts Team]: done (4 selections)
Selected analysts: market, social, news, fundamentals
╭──────────────────────────────────────────────────────────────────────────────╮
│                                                                              │
│  Step 4: Research Depth                                                      │
│  Select your research depth level                                            │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
? Select Your [Research Depth]: Deep - Comprehensive research, in depth debate a
nd strategy discussion
╭──────────────────────────────────────────────────────────────────────────────╮
│                                                                              │
│  Step 5: Thinking Agents                                                     │
│  Select your thinking agents for analysis                                    │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
? Select Your [Quick-Thinking LLM Engine]: GPT-4o - Standard model with solid ca
pabilities
? Select Your [Deep-Thinking LLM Engine]: o1 - Premier reasoning and problem-sol
ving model
╭────────────────────────── Welcome to TradingAgents ──────────────────────────╮
│                                                                              │`

saksham0161:
```
│                                                                              │
[...]

PR #28, opened 2025-06-11
Remove TSLA hardcoding from prompt

saksham0161:
https://github.com/TauricResearch/TradingAgents/issues/27


Yijia-Xiao:
Hi @saksham0161, that's a great catch! Thank you for the PR!

PR #30, opened 2025-06-14
fix: collection [bull_memory] already exists

minhphuc429:
### Problem
The application was throwing `Collection [bull_memory] already exists` error when trying to initialize `FinancialSituationMemory` instances on subsequent runs.

### Root Cause
The `FinancialSituationMemory` class was using `create_collection()` which fails if a collection with the same name already exists from previous runs.

### Solution
Changed `create_collection()` to `get_or_create_collection()` in the `FinancialSituationMemory.__init__()` method.

### Benefits
- ✅ Eliminates collection already exists error
- ✅ Enables persistent memory across application restarts
- ✅ Maintains backward compatibility
- ✅ No data loss from previous runs

### Files Changed
- `TradingAgents/tradingagents/agents/utils/memory.py`

### Testing
Verified that the application can now be restarted multiple times without encountering the ChromaDB collection error.

### DIFF
--- a/tradingagents/agents/utils/memory.py
+++ b/tradingagents/agents/utils/memory.py
@@ -1,20 +1,17 @@
 import chromadb
 from chromadb.config import Settings
 from openai import OpenAI
-import numpy as np
 
 
 class FinancialSituationMemory:
     def __init__(self, name):
         self.client = OpenAI()
         self.chroma_client = chromadb.Client(Settings(allow_reset=True))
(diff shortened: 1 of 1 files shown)


Yijia-Xiao:
Thank you @minhphuc429 for the contribution! The PR has some code conflicts that prevent direct merging, we will review soon and merge your PR.

Thanks!

Yijia-Xiao:
Thanks @minhphuc429! A more complete fix is tracked in #146.

PR #40, opened 2025-06-19
Updated README.md

RealMyth21:
I spent like 10 minutes wondering why it made no sense to me until I realized the diagrams were switched.


Yijia-Xiao:
Good point! The diagrams should be switched. Thanks, Mithil @RealMyth21.

PR #46, opened 2025-06-20
Updated requirements.txt based on latest commit

AtharvSabde:
PULL REQUEST: Add support for other backends, such as OpenRouter and Ollama

it had two requirements missing. added those


Yijia-Xiao:
Thanks @AtharvSabde for the contribution!

PR #48, opened 2025-06-22
feat: Initial FastAPI backend with auth and dummy strategies endpoint

QuantumLeap-us:
- Sets up a FastAPI application structure.
- Implements JWT-based authentication with /api/v1/auth/login endpoint.
- Adds a dummy /api/v1/strategies endpoint requiring authentication.
- Integrates SQLAlchemy with PostgreSQL and Alembic for user model migrations.
- Includes basic configuration, .env.example, and README updates for the new API.


George-Polya:
The official FastAPI documentation is written using SQLModel. I'm aware that SQLModel is built on top of SQLAlchemy, but which one is considered the better choice to use?

PR #98, opened 2025-07-04
set encoding to utf-8 when reading or writing files

Hellek1:
file access currently does not specify the encoding. this makes it ambigious and OS dependant, leading to random failures on some environments.

this PR sets the correct encoding (utf-8) explicitly, increasing the stability.


samchenku (approved):

Yijia-Xiao:
Thanks @Hellek1! Fixed in 929e041.

PR #116, opened 2025-07-07
feat: support for ollama users who run their models locally.

autotntfan:
### Motivation
Running TradingAgents on local Ollama models that lack the "tools" tag currently fails with a 400 error and OpenAI API.  This PR makes the framework automatically downgrade to plain reasoning so that fully open-source users can still run the main.py file. In addition, allow local Ollama model to be manually entried in CLI.

### What’s in This PR
* `utils.safe_bind_tools` helper (some ollama models don't have `.bind_tools` attribute)
* All analyst / researcher / trader nodes updated to use it (replace `.bind_tools` with `safe_bind_tools`)
* `utils.memory` embedding model is updated (if ollama model is used, the original OpenAI model is replaced with a custom model)
* `cli/utils.py` added local ollama model selections, where user can input their model.
* `cli/main.py`  added ollama model for embedding to avoid errors raised by OpenAI API.

### Test Models Used in `main.py`
| model:tag                                    | tools attached | status |
|------------------------------------|----------------|--------|
| fingpt:7b                                              | ❌ no         | works |
| mixtral:8x7b-instruct-v0.1-q4_K_M`     | ❌ no         | works |
| phi3:latest                                            | ❌ no         | works |
| mistral-nemo:latest                             | ✅ yes        | works |

### CLI Test
```
Select your LLM Provider: 
Ollama

Select Your [Quick-Thinking LLM Engine] LLM Engine: 
Custom model (type manually)

Enter the exact Ollama model name for Quick-Thinking LLM Engine: 
phi3:mini

Select Your [Deep-Thinking LLM Engine] LLM Engine: 
Custom model (type manually)

Enter the exact Ollama model name for Deep-Thinking LLM Engine: 
mistral-nemo:latest

Select Your [Embedding LLM Engine] LLM Engine:
Custom model (type manually)

Enter the exact Ollama model name for Embedding LLM Engine:
fingpt:7b
```

### Backward Compatibility
Ollama backend url is changed in `BASE_URLS` @ cli/utils.py/select_llm_provider
`/v1` in the original url is removed. If any error arises, restore `/v1` in the tail of this backend url.

### My Platform
* Windows 10 (WSL v.2.5.9.0)
* CPU: Intel 13600k
* GPU: Nvidia RTX 4070 Ti Super 16GB
* RAM: DDR4 3200 64 GB

### Example
python main.py
python -m cli.main

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -479,7 +479,8 @@ def get_user_selections():
     )
     selected_shallow_thinker = select_shallow_thinking_agent(selected_llm_provider)
     selected_deep_thinker = select_deep_thinking_agent(selected_llm_provider)
-
+    selected_embedding_model = select_embedding_agent(selected_llm_provider)
+    
     return {
         "ticker": selected_ticker,
         "analysis_date": analysis_date,
@@ -489,6 +490,7 @@ def get_user_selections():
         "backend_url": backend_url,
         "shallow_thinker": selected_shallow_thinker,
         "deep_thinker": selected_deep_thinker,
+        "embedding_model": selected_embedding_model,
     }
 
 
@@ -741,6 +743,7 @@ def run_analysis():
     config["max_risk_discuss_rounds"] = selections["research_depth"]
     config["quick_think_llm"] = selections["shallow_thinker"]
     config["deep_think_llm"] = selections["deep_thinker"]
+    config["embedding_model"] = selections["embedding_model"]
     config["backend_url"] = selections["backend_url"]
     config["llm_provider"] = selections["llm_provider"].lower()
 
--- a/cli/utils.py
+++ b/cli/utils.py
@@ -1,5 +1,5 @@
 import questionary
-from typing import List, Optional, Tuple, Dict
+from typing import List, Optional, Tuple, Dict, Sequence
 
 from cli.models import AnalystType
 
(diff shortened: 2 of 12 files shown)


Matthewlo213:
Is there a tutorial on using the Ollama local model to avoid the OpenAI API key? I am stuck in this, and I can't run the program

autotntfan:
> Is there a tutorial on using the Ollama local model to avoid the OpenAI API key? I am stuck in this, and I can't run the program

Apologies, but when I ran `main.py` with Ollama models on both Windows and WSL, I didn’t encounter any “OpenAI API Key” errors. ( `export OPENAI_API_KEY=key` cannot be used on Windows, so the key is probably not set yet)

**Additional note on how I run the project**

After specifying the models in `main.py`, I simply execute the script from the `TradingAgents/` directory:
(you can directly run main.py in this commit)

```bash
user\TradingAgents > pip install langchain_ollama
user\TradingAgents > python main.py
```

Alternatively, you can launch the interactive CLI:

```bash
python -m cli.main
```

and follow the on-screen prompts to choose the models.


BTW, You might try registering for a key on the OpenAI website first, then set it in your shell:

```bash
export OPENAI_API_KEY=your_key_here
```

and see whether that solves the problem on your end.

Matthewlo213:
What if my region cannot register for an OpenAI API key, and I set the OpenAI API key to NULL? Would there be a problem while getting the news from a specific stock? I have solved the problem of running the program. Thank you.

autotntfan:
> What if my region cannot register for an OpenAI API key, and I set the OpenAI API key to NULL? Would there be a problem while getting the news from a specific stock? I have solved the problem of running the program. Thank you.

You cannot literally set Open API key to NULL. Any helper whose name ends in _openai (e.g. get_stock_news_openai, get_fundamentals_openai) will call the OpenAI SDK, and if the key is missing the SDK returns 401 “key required”. 

Until now those helpers were never reached when you ran an Ollama model with the tool tag, because a bug in `safe_bind_tools.py` prevented the toolkit from being attached. I’ve fixed that in the latest commit:

- Tool-enabled models such as qwen3 (quick) or mixtral (deep) now correctly execute the toolkit.
- For `social` analysis. When `get_stock_news_openai` (inside the social ToolNode) is called without a key it throws an auth error, but the error is caught, so the run continues—just without social sentiment data.
image

- Same for `news` analysis: it tries `get_global_news_openai`, catches the auth error, then falls back to `get_google_news`.

image

PS:
If you want to see exactly which toolkit functions are invoked for each analysis step, open `trading_graph.py`—the `_create_tool_nodes` method lists every ToolNode (market, social, news, fundamentals) and the helper functions it tries, in the order they’re called.

So the agent finishes, but if you have no OPENAI_API_KEY it ignores the OpenAI-powered signals and works only with Google / FinMind data.

Hope that clarifies things, and thanks again for your helps!
Please let me know if you have any questions.~

arpieb:
I'm also interested in running this against a locally-hosted Ollama server that I built out to specifically address the privacy + cost concerns re OpenAI.

Yijia-Xiao:
Thanks @autotntfan for the early local-Ollama proposal. TradingAgents has shipped native Ollama support since v0.2.x (pickable in the CLI dropdown), and remote Ollama endpoints are now configurable via `OLLAMA_BASE_URL` as of f10daa2. Closing as superseded.

PR #117, opened 2025-07-07
feat: allow custom model and API URL input for LLM selection in CLI u…

opqrst-chen:
…tilities


Yijia-Xiao:
Thanks @opqrst-chen! TradingAgents already supports any OpenAI-compatible endpoint via the existing provider config: set llm_provider to one of our OpenAI-compatible presets (or use the OpenRouter entry) and override base_url + model_name in the CLI or in config. Closing as already-supported, with appreciation for raising the use case.

PR #120, opened 2025-07-08
support get price data from alpha vantage

auwc:
Since yahoo finance always hit rate limit, replace stock price data acquirement procedure with alpha vantage could solve this problem.
The api key is easy to apply and free.


Tauric-Research:

PR #159, opened 2025-07-20
feat: Save Risk Management Team Debate's output as separate markdown file

thumlx12:
add a new key in `report_section` named 'risk_debate' to store Risk Management Team Debate output into a separate markdown file.


Tauric-Research:

PR #200, opened 2025-08-15
feat: custom provider and model

mogita:
**Changes:**

- [x] Added dotenv to auto-load `.env`
- [x] Added custom provider support via `CUSTOM_BASE_URL` and `CUSTOM_API_KEY` environment variables. The custom providers should provide OpenAI-compatible interface.
- [x] Custom provider appears as "Custom (hostname)" option first in provider selection list
- [x] Added "Custom Model" fallback option for manual model name input

**Updated workflow:** Users can now set `CUSTOM_BASE_URL` and `CUSTOM_API_KEY` in `.env` to add a custom OpenAI-compatible provider option. When selected, users choose from a unified list of all available models plus a "Custom Model" option for manual input. The system uses OpenAI's ChatOpenAI client for all custom provider requests, thus requiring endpoints to be OpenAI-compatible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ eval_results/
 eval_data/
 *.egg-info/
 .env
+*.log
+results/*
--- a/cli/main.py
+++ b/cli/main.py
@@ -19,6 +19,10 @@ from rich.tree import Tree
 from rich import box
 from rich.align import Align
 from rich.rule import Rule
+from dotenv import load_dotenv
+
+# Load environment variables from .env file
+load_dotenv()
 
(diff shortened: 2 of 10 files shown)


porterlock112 (approved):

Yijia-Xiao:
Thanks @mogita! TradingAgents already supports any OpenAI-compatible endpoint via the existing provider config: set llm_provider to one of our OpenAI-compatible presets (or use the OpenRouter entry) and override base_url + model_name in the CLI or in config. Closing as already-supported, with appreciation for raising the use case.

PR #207, opened 2025-08-24
Ensure all debate state fields initialized

Mgiri1234:


Yijia-Xiao:
Thanks @Mgiri1234! This was addressed in eec6ca4.

PR #217, opened 2025-09-05
[Bug] Fix Stock Data Handling to Prevent Calculation Errors

samchenku:
This PR addresses issues with corrupted or malformed stock data in CSV files, which previously caused calculation errors in stock indicators. Specifically, it:

1. Adds on_bad_lines='skip' when reading CSVs to prevent crashes from malformed lines.
2. Ensures date parsing handles invalid or corrupted dates with errors='coerce' and removes rows with missing dates.
3. Cleans NaN values in price columns before calculating stock indicators, including forward/backward filling to handle stubborn missing values.
4. Improves robustness of functions that calculate stock stats, avoiding errors like Cannot mask with non-boolean array containing NA / NaN values.

This ensures stock statistics are calculated reliably even if source CSVs contain bad data.

### DIFF
--- a/tradingagents/dataflows/interface.py
+++ b/tradingagents/dataflows/interface.py
@@ -517,7 +517,9 @@ def get_stock_stats_indicators_window(
             os.path.join(
                 DATA_DIR,
                 f"market_data/price_data/{symbol}-YFin-data-2015-01-01-2025-03-25.csv",
-            )
+            ),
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @samchenku! Applied a systematic fix in b0f9d18 with a shared data cleaning pipeline across both stockstats_utils.py and y_finance.py.

PR #223, opened 2025-09-12
提交AI模型使用阿里qwen的版本

OliverPanda:


OliverPanda:
测试了能用qwen跑通

abhishekgahlot2:
Would it not make sense to use the litellm adapter here to so other models can be implemented or used quickly? https://github.com/BerriAI/litellm

Yijia-Xiao:
Thanks @OliverPanda for the early Qwen integration proposal. TradingAgents now ships native Qwen support with dual-region endpoints (faaeeba): international via dashscope-intl.aliyuncs.com (`DASHSCOPE_API_KEY`) and China via dashscope.aliyuncs.com (`DASHSCOPE_CN_API_KEY`), selectable via a secondary region prompt in the CLI. Closing as superseded.

PR #252, opened 2025-10-16
Add vLLM OpenAI-Compatible API Support

reopio:
### Summary
This PR adds support for vLLM's OpenAI-compatible API, allowing users to leverage vLLM as an alternative LLM backend.

### Changes
- Added vLLM as a new LLM provider option
- Implemented configuration prompts for vLLM setup:
  - Custom vLLM API URL input
  - Model name specification
  - Embedding model name specification
- Integrated vLLM with OpenAI-compatible API interface

### Usage
When selecting vLLM as the LLM provider, users will be prompted to manually enter:
1. **API URL**: The custom vLLM API endpoint (e.g., `http://localhost:8000/v1`)
2. **Model Name**: The name of the LLM model to use
3. **Embedding Model Name**: The name of the embedding model to use

### Benefits
- Provides flexibility to use self-hosted vLLM instances
- Maintains compatibility with OpenAI API standards
- Enables users to leverage vLLM's high-performance inference capabilities

### Testing
-  Verified vLLM API connection with custom URL
-  Tested model inference functionality
-  Validated embedding generation

### DIFF
--- a/cli/utils.py
+++ b/cli/utils.py
@@ -152,24 +152,43 @@ def select_shallow_thinking_agent(provider) -> str:
         "ollama": [
             ("llama3.1 local", "llama3.1"),
             ("llama3.2 local", "llama3.2"),
+        ],
+        "vllm": [
+            ("llama3.1 local", "llama3.1"),
+            ("qwen3", "qwen3"),
         ]
     }
 
-    choice = questionary.select(
-        "Select Your [Quick-Thinking LLM Engine]:",
-        choices=[
-            questionary.Choice(display, value=value)
-            for display, value in SHALLOW_AGENT_OPTIONS[provider.lower()]
-        ],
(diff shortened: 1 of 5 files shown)


ZhengHongming888:
This is one good PR which can use self-hosting endpoint instead of costive openai API. Suggest to merge asap!!! I can help to test.

Yijia-Xiao:
Thanks @reopio for the vLLM proposal. vLLM exposes an OpenAI-compatible HTTP server, so it works through the existing flow today: pick the `OpenAI` provider in the CLI and set `TRADINGAGENTS_LLM_BACKEND_URL=http://your-vllm-host:8000/v1` (with whatever key your vLLM deployment expects as `OPENAI_API_KEY`). Env-var support shipped in d13e9b7. Closing as redundant.

PR #258, opened 2025-10-22
XAI support and using Hugging Face for embeddings and support multiple language

jeffreychuuu:
- add support for using XAI
- using Hugging Face for embeddings as XAI does not support embedding natively
- support Traditional Chinese, Simplified Chinese and English


Yijia-Xiao:
Thanks @jeffreychuuu for the bundled XAI + Hugging Face embeddings + multi-language proposal. Two of the three pieces have shipped: native xAI is in the provider list (uses XAI_API_KEY), and `output_language` now propagates to every user-facing agent (6b384f7, #575). The Hugging Face embeddings piece is the part that doesn't yet have a home — if it's still important to you, a focused PR scoped just to embeddings would be much easier to review than this bundle. Closing this PR as partially superseded; happy to look at a smaller follow-up.

PR #265, opened 2025-10-28
fix: uv package issue and we now are able to install tradingagents by uv

Jasonsey:
fix uv package issue and we now are able to install tradingagents by uv


Yijia-Xiao:
Thanks @Jasonsey! Package installation has been restructured with pyproject.toml as the single source of truth. Feel free to reopen if you encounter installation issues.

Issue #275, opened 2025-11-09
Gemini & OpenRouter doesnt work


zigzag-way:
I tried all combinations in the shallow mode (TLT, current date, Shallow, Openrouter\Gemini) but every time it failed for different reasons, like 
- RuntimeError: All vendor implementations failed for method 'get_global_news'
- RuntimeError: All vendor implementations failed for method 'get_indicators'
- 403 and others...

GEMINI & ALPHAVANTAGE keys are set in .env

Tauric-Research:
Thanks for the report. We could not reproduce this on the current version, so closing for now; reopen with the version, the command you ran and the error output, and we will take another look.

PR #276, opened 2025-11-14
Fix #275: Make openai dataflow compatible with Gemini and OpenRouter

godnight10061:
- Replace OpenAI-specific responses.create() API with standard chat.completions.create()
- Update get_stock_news_openai, get_global_news_openai, and get_fundamentals_openai
- Add comprehensive tests for OpenAI, Gemini, and OpenRouter compatibility
- All functions now use standard OpenAI-compatible chat completion API
- Fixes RuntimeError: All vendor implementations failed for method 'get_global_news'
- Fixes RuntimeError: All vendor implementations failed for method 'get_indicators'

The issue was that the openai vendor functions used OpenAI-specific API features (responses.create with web_search_preview tools) that are not supported by Gemini or OpenRouter. By switching to the standard chat completions API, these functions now work with any OpenAI-compatible provider.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,2 @@
+# Tests for TradingAgents
+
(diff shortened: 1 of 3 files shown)


Yijia-Xiao:
Shipped — responses are normalized to strings (handles Gemini's typed-block lists) and the OpenAI base URL no longer leaks into other providers. Closing as superseded.

Issue #274, opened 2025-11-09
get_global_news_openai not getting latest news


samchenku:
Openai is hallucinating and provides outdated news.

Example:
```
1. **U.S. Dollar Decline and European Stock Rally**  
   - On November 6, the U.S. dollar weakened significantly, down by as much as 1.3% against major currencies including the euro, Canadian dollar, Mexican peso, and offshore Chinese yuan.  
   - European equity markets benefited, with the STOXX 600 index up by 0.3% and the MSCI All-World index rising by 0.4%.  
   - This rally was linked to reports that President-elect Donald Trump would delay implementing some proposed tariffs, easing some trade tensions temporarily.

2. **China's Large Fiscal Stimulus**  
   - On November 8, China announced a fiscal stimulus package totaling approximately $1.4 trillion (10 trillion yuan).  
   - Key measures include increasing the local government debt ceiling by 6 trillion yuan and issuing 800 billion yuan in special local government bonds annually for five years.  
   - The stimulus aims at reducing local government indebtedness and boosting economic growth amid global uncertainties.

3. **Upcoming G20 Johannesburg Summit (Nov 22-23, 2025)**  
   - The focus will be on solidarity, equality, and sustainable development, with a strong emphasis from South Africa on Africa’s development and global debt reform.  
   - This summit is likely to influence global cooperation frameworks and may impact emerging market economic policies.

4. **U.S. Tariff Policy and Trade Tensions**  
   - Early November saw the imposition by President Trump of a 100% tariff on Chinese goods effective Nov 1, in response to China’s export controls on rare earth minerals.  
   - An additional 10% tariff was imposed on Canadian goods, citing political issues in Ontario opposed to U.S. tariffs.  
   - These escalating tariffs contribute to ongoing trade tensions, adding volatility to global markets.

5. **Market Volatility Amid Trade Uncertainty**  
   - The S&P 500 has experienced significant swings reflecting investor caution.  
   - Asian markets like the Hang Seng and Nikkei 225 underperformed due to trade worries and economic uncertainties.  
   - U.S. 10-year Treasury yields rose to 4.18%, indicating inflation concerns related to trade and tariff policies.
```

Tauric-Research:
Thanks for the report. This is addressed in the current version, so closing as fixed; if you still see it on the latest main, reopen with the version and a short repro.

PR #277, opened 2025-11-14
Fix #274: Add warnings for OpenAI news/fundamentals hallucination risk

godnight10061:
- Add UserWarning to get_global_news_openai, get_stock_news_openai, and get_fundamentals_openai
- Warnings inform users that OpenAI vendor may hallucinate or provide outdated data
- Recommend alternative vendors: alpha_vantage, google, yfinance, or local
- Add comprehensive test suite to verify warnings are emitted
- Refactor to use shared _warn_hallucination_risk helper function

The issue reported that OpenAI was hallucinating and providing outdated news (e.g., fake news from November 2025). This is because the OpenAI vendor relies on the LLM's training data rather than real-time web search. Users should use alternative vendors for reliable, up-to-date news and fundamental data.

All tests passing (4/4).

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,2 @@
+# Tests package
+
--- a/tests/test_openai_news_warnings.py
+++ b/tests/test_openai_news_warnings.py
@@ -1,0 +1,124 @@
+"""
+Tests for OpenAI news dataflow functions to ensure proper warnings about hallucination risks.
+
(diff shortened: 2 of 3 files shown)


Yijia-Xiao:
Addressed at the source — the analyzed identity is resolved deterministically (#814) and the market analyst grounds exact numbers in a verified snapshot (#830). Closing as superseded.

PR #278, opened 2025-11-15
add new mode 'openai_compatible'

grandgen:
Completed all the necessary code changes to support Google's Gemini model via an OpenAI-compatible endpoint. Here is a summary of the changes:

trading_graph.py: Updated the LLM instantiation logic to include a new "openai_compatible" provider. This new provider uses the ChatOpenAI class and is configured using the OPENAI_COMPATIBLE_API_KEY and OPENAI_COMPATIBLE_BASE_URL environment variables.
default_config.py: Updated the default configuration to include the new "openai_compatible" provider and the associated settings.
utils.py: Added "OpenAI Compatible" to the list of LLM providers in the CLI and added a placeholder for the Gemini Pro model.
README.md: Updated the documentation to instruct users on how to configure the system for Vertex AI and other OpenAI-compatible endpoints.
You can now configure the project to use your Gemini API key by setting the appropriate environment variables in your .env file and selecting the "OpenAI Compatible" provider when running the CLI.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,3 @@
 ALPHA_VANTAGE_API_KEY=alpha_vantage_api_key_placeholder
 OPENAI_API_KEY=openai_api_key_placeholder
+GOOGLE_API_KEY=google_api_key_placeholder
--- a/README.md
+++ b/README.md
@@ -114,11 +114,15 @@ pip install -r requirements.txt
 
 ### Required APIs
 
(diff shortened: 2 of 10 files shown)


Yijia-Xiao:
Thanks @grandgen! TradingAgents already supports any OpenAI-compatible endpoint via the existing provider config: set llm_provider to one of our OpenAI-compatible presets (or use the OpenRouter entry) and override base_url + model_name in the CLI or in config. Closing as already-supported, with appreciation for raising the use case.

PR #282, opened 2025-11-19
Turns-off upload *.* files, addresses conflicts in previous PR

jackspace:


Yijia-Xiao:
Thanks @jackspace for the follow-up! Closing alongside #281 for the same reason: scope. Focused PRs for individual subsystems would be much easier to land.

PR #286, opened 2025-11-20
Fix/openrouter embeddings

00make:


Yijia-Xiao:
Embeddings were removed (the per-agent embedding memory became a persistent decision log), so the OpenRouter embeddings path no longer exists. Closing as superseded.

PR #290, opened 2025-11-24
kFix: Stop after all primary vendors in multi-vendor config

Aitous:
When multiple primary vendors are configured (e.g., 'reddit,alpha_vantage'), the system now correctly stops after attempting all primary vendors instead of continuing through all fallback vendors.

Changes:
- Track which primary vendors have been attempted in a list
- Add stopping condition when all primary vendors are attempted
- Preserve existing single-vendor behavior (stop after first success)

This prevents unnecessary API calls and ensures predictable behavior.


Yijia-Xiao:
The vendor routing has been refactored and this behavior is now handled correctly. Thanks @Aitous for the contribution# Also close #290 (related, same area)

PR #328, opened 2026-01-30
New UI

qq173681019:


Yijia-Xiao:
Thanks @qq173681019 for the UI proposal. TradingAgents stays CLI-focused; UI work fits better as a separate companion project. Closing here.

PR #355, opened 2026-03-02
Add support for accessing LLM models on Azure Foundry

yulinzhang96:
## Summary

- This change was created by Claude Opus 4.6, and run on Windows x64 platform.
- Azure Foundry supports any models in the catalog, so there's no need to select or switch to different LLM providers; users can therefore deploy any models they want and use the pricing on Azure Foundry.
- The change adds a new client for interacting with Azure Foundry, and updates the usage in appropriate places.
- Users should set their Foundry project endpoint in tradingagents/default_config.py, the format of which is `https://<resource>.openai.azure.com/openai/v1/`; they should also set `AZURE_FOUNDRY_API_KEY` in the environment variable accordingly.
- Fix data handling and file I/O exceptions when running the program on Windows by adding `encoding="utf-8"`.

## References
- https://azure.microsoft.com/en-us/products/ai-foundry


Yijia-Xiao:
Thanks @yulinzhang96 for the contribution! Azure support implemented in b0f6058.

PR #360, opened 2026-03-06
add DeepSeek support

null0NULL123:


Yijia-Xiao:
Thanks @null0NULL123! DeepSeek support added in b0f6058.

PR #361, opened 2026-03-06
fix: pass debate round config to ConditionalLogic

makk9:
ConditionalLogic was always initialized with default values (1 round each), ignoring max_debate_rounds and max_risk_discuss_rounds from user config. This meant users had no way to control debate iterations.


Yijia-Xiao:
Good catch @makk9 !

11591599 (commented):
Evva

PR #371, opened 2026-03-10
add Simplified Chinese option for reports

divingken:


Yijia-Xiao:
Thanks @divingken! Multi-language output support (including Simplified Chinese) is now available via the CLI language selector. Feel free to reopen if anything was missed.

PR #324, opened 2026-01-17
feat: multi-provider support, retry logic, and dynamic model fetching

MUmarJ:
## Summary

This PR adds several improvements to enhance multi-LLM provider support, error handling, and CLI usability.

### 1. Content Normalization for Multi-Provider Support
- Added `normalize_content()` utility to handle Gemini's list-of-dicts response format alongside OpenAI/Anthropic strings
- Applied normalization in all analyst modules (market, news, fundamentals, social media)

### 2. Retry Logic with Exponential Backoff
- Added retry logic with exponential backoff (1s, 2s, 4s) for transient network errors
- Smart error categorization: retries for connection/timeout errors, immediate vendor fallback for rate limits
- Better logging with RETRY/TRANSIENT_ERROR/RATE_LIMIT states

### 3. OpenAI Responses API Support (GPT-5 Ready)
- New `tradingagents/llm/` module with LangChain-compatible Responses API wrapper
- Auto-detection of models requiring Responses API (`gpt-5*`, `codex*`)
- Dynamic `web_search` tool type selection based on `FETCH_LATEST` env var
- Robust response text extraction that handles varying output structures

### 4. Configuration Validation for Local LLM Providers
- Added `validate_config()` that warns when using local LLM with OpenAI-only data vendors
- LM Studio added to supported providers list
- Actionable recommendations for incompatible configurations

### 5. Dynamic Model Fetching & CLI Enhancements
- Dynamic model fetching from OpenAI, Anthropic, Google APIs when `FETCH_LATEST=true`
- Provider availability checking with health probes for local providers (Ollama, LM Studio)
- Updated static model lists with latest models (Gemini 2.5/3, new OpenRouter models)
- Separate "Quick-Thinking" and "Deep-Thinking" model selection steps
- Disabled options shown with reasons ("No API key", "Not running")

## Issues Addressed

- Fixes #275 - Gemini & OpenRouter compatibility issues
- Helps with #296 - Multi-LLM provider support
- Fixes #289 - Fallback vendor logic improvements
- Helps with #305 - Rate limiting handling
- Fixes #317 - Gemini 2.5 Flash Preview model error
- Helps with #274 - Web search improvements

PR #459, opened 2026-03-25
feat: enhance LM Studio integration and configuration

jhicken:
- Updated .env.example to include API keys and configuration for various LLM providers, including LM Studio.
- Modified cli/utils.py to implement fetch_lmstudio_models function for dynamic model retrieval.
- Updated cli/main.py and cli/utils.py to support LM Studio as a provider option.
- Enhanced tradingagents/llm_clients to accommodate LM Studio in client creation and API key handling.
- Updated .gitignore to exclude results and reports directories. (this get my commits all messy)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,52 @@
-# LLM Providers (set the one you use)
+# =============================================================================
+# TradingAgents Environment Configuration
+# =============================================================================
+# Copy this file to .env and fill in your API keys and configuration
+
+# =============================================================================
+# LLM Provider API Keys
+# =============================================================================
+# Set the API key for the LLM provider you want to use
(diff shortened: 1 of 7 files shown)


jhicken:
I missed this #324 when i put this together. I don't have any particular attachment to my implementation. Just so long as it works.

Yijia-Xiao:
Thanks @jhicken for the LM Studio integration work! LM Studio's OpenAI-compatible endpoint is already reachable through the existing Ollama provider entry by setting a custom base URL, so a dedicated provider class is not required. Closing in favor of that path; if specific LM Studio behaviors still feel missing, a focused issue describing them would be most helpful.

PR #461, opened 2026-03-25
Add Cerebras as LLM provider

fefe2673:
## Summary
- Add Cerebras provider support in openai_client.py (base URL + API key config)
- Register cerebras in factory.py so it routes through the OpenAI-compatible client
- Update main.py to use Cerebras with llama-4-scout-17b-16e-instruct model
- Add CEREBRAS_API_KEY to .env.example

## Test plan
- [ ] Set CEREBRAS_API_KEY in .env and run python main.py
- [ ] Verify the trading agent graph completes with Cerebras as the LLM provider

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -927,6 +927,10 @@ def run_analysis():
     config["deep_think_llm"] = selections["deep_thinker"]
     config["backend_url"] = selections["backend_url"]
     config["llm_provider"] = selections["llm_provider"].lower()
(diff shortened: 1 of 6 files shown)


fefe2673:
n

Yijia-Xiao:
Thanks @fefe2673! Cerebras and similar providers can be used via OpenRouter. DeepSeek, Qwen, and GLM added as first-class providers in b0f6058.

PR #469, opened 2026-03-26
Support for Perplexity API Key

santidediego:
Added support for Perplexity API Key. Tested with my own key.


Yijia-Xiao:
Thanks @santidediego for the Perplexity proposal. Perplexity exposes an OpenAI-compatible API at `https://api.perplexity.ai`, so it works through the existing flow: pick `OpenAI` and set `TRADINGAGENTS_LLM_BACKEND_URL=https://api.perplexity.ai` (with your Perplexity key as `OPENAI_API_KEY`). Shipped d13e9b7. Closing as redundant.

PR #488, opened 2026-03-31
fix(dataflows): add missing pandas import in y_finance.py

JasonOA888:
Fix a missing `import pandas as pd` in `y_finance.py`.

The `_get_stock_stats_bulk()` function uses `pd.isna()` on line 213 but pandas was never imported in this file. It worked by accident because `stockstats_utils` was imported at module level and loaded pandas into the process namespace. This is a fragile implicit dependency that breaks if import order changes.


Yijia-Xiao:
Thanks @JasonOA888 for the contribution! Fixed in 28d5cc6.

PR #489, opened 2026-03-31
Fix missing pandas import in y_finance bulk indicators

samchenku:
## Summary
- add the missing `pandas` import used by `_get_stock_stats_bulk()` in `tradingagents/dataflows/y_finance.py`
- add a regression test covering bulk indicator handling when the computed value is `NaN`

## Root cause
`_get_stock_stats_bulk()` calls `pd.isna(...)`, but `pandas` was never imported in `y_finance.py`. That causes a `NameError` when bulk indicator results contain missing values.

## Impact
This fixes bulk stockstats indicator lookups so missing indicator values are converted to `"N/A"` instead of crashing.

## Testing
- `PYTHONPATH=/tmp/tradingagents-upstream-pr pytest -q tests/test_y_finance_bulk_indicator.py`

### DIFF
--- a/tests/test_y_finance_bulk_indicator.py
+++ b/tests/test_y_finance_bulk_indicator.py
@@ -1,0 +1,39 @@
+import importlib
+import sys
+import types
+import unittest
+from unittest.mock import patch
+
+import pandas as pd
+
+
+class YFinanceBulkIndicatorTests(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @samchenku for the contribution! Fixed in 28d5cc6.

PR #490, opened 2026-03-31
Normalize technical indicator names in tool routing

samchenku:
## Summary
- normalize indicator names to lowercase before routing them to the configured vendor
- support comma-separated indicator requests in `get_indicators`
- add a regression test covering `RSI, MACD` input

## Root cause
`get_indicators` forwarded the raw user-provided indicator string directly to the vendor layer. Inputs like `RSI` or `RSI, MACD` failed because the downstream yfinance/stockstats path expects lowercase indicator keys such as `rsi` and `macd`.

## Testing
- `python -m unittest discover -s tests -p 'test_technical_indicators_tools.py'`


Yijia-Xiao:
Thanks @samchenku for the contribution! Fixed in 78fb66a.

PR #533, opened 2026-04-12
fix(tests): stub missing dependencies

Gujiassh:
## Summary
- add lightweight stubs for langchain, pandas, stockstats, yfinance, rank_bm25, and questionary so the repo can be imported without installing every dependency
- ensure tests insert the repo root into  so  and  modules resolve during pytest runs
- stub the necessary classes/methods used by the validator and CLI tests

## Testing
- pytest

### DIFF
--- a/langchain_anthropic.py
+++ b/langchain_anthropic.py
@@ -1,0 +1,6 @@
+class ChatAnthropic:
+    def __init__(self, **kwargs):
+        self.kwargs = kwargs
+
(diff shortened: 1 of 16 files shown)


Yijia-Xiao:
Thanks @Gujiassh for the contribution! Our pytest fixtures in tests/conftest.py now provide placeholder API keys and lazy-load LLM clients, so the test suite no longer needs dependency stubs. Closing as covered.

PR #534, opened 2026-04-12
fix(cli): log all tool calls from streamed chunk messages

### LINKED ISSUE #520: Bug report (possibly)
On multiple tool calls in one turn, only the result of the last tool call is saved to the log file, i.e. message_tool.log, while all tool calls are correctly executed and returned. It seems to be a bug.
Here is an example:
Image

Gujiassh:
## Summary
- fix streamed message ingestion to process all messages in each chunk instead of only `chunk["messages"][-1]`
- deduplicate by message id while still recording every tool call for newly seen messages
- add focused unit tests for multi-tool-call ingestion and duplicate-id suppression

## Problem
Issue #520 reports that `message_tool.log` only keeps the last tool call when multiple tool calls occur in one turn.

The root cause is in CLI stream handling: it only inspects the last message in each chunk, so earlier messages/tool calls in the same chunk are ignored.

## Changes
- added `cli/chunk_ingest.py` with `ingest_chunk_messages(...)`
- updated `cli/main.py` to call the new helper in the stream loop
- added `tests/test_cli_chunk_message_ingest.py`

## Validation
- `PYTHONPATH=. pytest -q tests/test_cli_chunk_message_ingest.py`

Closes #520

### DIFF
--- a/cli/chunk_ingest.py
+++ b/cli/chunk_ingest.py
@@ -1,0 +1,19 @@
+def ingest_chunk_messages(message_buffer, chunk, classify_message_type) -> None:
+    """Ingest all newly seen messages from a graph stream chunk."""
+    for message in chunk.get("messages", []):
+        msg_id = getattr(message, "id", None)
+        if msg_id is not None:
+            if msg_id in message_buffer._processed_message_ids:
+                continue
+            message_buffer._processed_message_ids.add(msg_id)
+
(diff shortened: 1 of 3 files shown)


Yijia-Xiao:
Thanks @Gujiassh for the contribution! Fixed in fa4d01c.

PR #546, opened 2026-04-16
feat: add MiniMax provider support

jyek:
## Summary

- Adds MiniMax as a natively supported LLM provider via its OpenAI-compatible API
- Adds model catalog entries for MiniMax-M2.7, MiniMax-M2.7-highspeed, MiniMax-M2.5, and MiniMax-M2.5-highspeed
- Adds MINIMAX_API_KEY to .env.example
- Adds retry handling for null choices responses caused by MiniMax content moderation

## Test plan

- [ ] Set MINIMAX_API_KEY in .env
- [ ] Run tradingagents run, select MiniMax as provider
- [ ] Verify all four models appear in the model selection list
- [ ] Confirm a full analysis run completes successfully

Generated with [Claude Code](https://claude.com/claude-code)


malaccan:
will this minimax support be pushed to main soon?

Yijia-Xiao:
Thanks @jyek! MiniMax now built-in as of 19d22b5, scaled to cover both regional endpoints (global and China) since contributors converged on that need.

PR #547, opened 2026-04-16
fix(cli): log all tool calls from streamed chunk messages

Gujiassh:
## Summary
- ingest every newly streamed message in a graph chunk instead of only the last entry
- record tool calls from each streamed message while still deduplicating repeated message ids
- cover the multi-message/multi-tool-call path with focused CLI ingestion tests

## Testing
- python3 -m pytest tests/test_cli_chunk_message_ingest.py -q
- python3 -m py_compile cli/main.py cli/chunk_ingest.py tests/test_cli_chunk_message_ingest.py


Yijia-Xiao:
Thanks @Gujiassh for the contribution! The same fix landed via #534. Closing as already covered.

PR #549, opened 2026-04-16
Fix Unicode encoding error in report file writes

### LINKED ISSUE #543: Error saving report: 'charmap' codec can't encode character '\u2248' in position 5153: character maps to <undefined> "
Error received when trying to save report : "Error saving report: 'charmap' codec can't encode character '\u2248' in position 5153: character maps to <undefined> "

atharvajoshi01:
Fixes #543

The report saving in `cli/main.py` opens files without specifying encoding, which defaults to the system locale on Windows. This causes a `charmap` codec error when reports contain Unicode characters like `≈` (\u2248).

Added `encoding="utf-8"` to the `open()` call so reports save correctly regardless of platform.

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -1005,7 +1005,7 @@ def run_analysis():
                 if content:
                     file_name = f"{section_name}.md"
                     text = "\n".join(str(item) for item in content) if isinstance(content, list) else content
-                    with open(report_dir / file_name, "w") as f:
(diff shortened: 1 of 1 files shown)


Yijia-Xiao:
Thanks @atharvajoshi01 for the contribution! Addressed in 872b063 with explicit encoding="utf-8" on every open() and write_text() call across the CLI and graph modules.

Issue #82, opened 2025-07-02
How about crypto trading？


ArbifyFun:
Maybe we can use it for crypto trading. Welcome all interested friends to join the discussion.

BlockchainsFans:
good ider 🐱‍🏍

Yung-Chih-Lo:
I will do it soon~

ArbifyFun:
> I will do it soon~

That‘s great ！

Zeyu-yeqiong:
I tried entering BTC directly and it work

ChengYen-Tang:
Hi @Yung-Chih-Lo,

For Crypto, I recommend several reference sources for Agent crawlers to refer to and analyze.

1. Cointelegraph
    - [BTC](https://cointelegraph.com/tags/bitcoin)
    - [ETH](https://cointelegraph.com/tags/ethereum)
2. [Glassnode Insights](https://insights.glassnode.com/tag/newsletter/)
    - Mainly insights from on-chain data
3. [交易所鏈上資金流動](https://www.coincarp.com/zh-tw/exchangeflow/)
4. https://www.coinglass.com/zh/newsflash
5. https://truthsocial.com/@realDonaldTrump  😂

ChengYen-Tang:
It can help us understand the project structure more easily

https://deepwiki.com/TauricResearch/TradingAgents/1-tradingagents-overview

xiaobanni:
Follow the thread.

lucasa:
Me too, starting now!

echoyinke:
![Image](https://github.com/user-attachments/assets/e7ee7096-5041-4c00-b603-57883d1001e1)

@lucasa @ArbifyFun @ChengYen-Tang @xiaobanni @Zeyu-yeqiong 





> good ider 🐱‍🏍



> I will do it soon~

Maybe join the group to dream big~

CadeYu:
I think crypto support is worth discussing, but it may be better to scope it as an analysis-first feature rather than a full trading/execution feature.

From the current codebase, TradingAgents is still very stock/company-centric:
- prompts and state use `company_of_interest`
- the analyst flow assumes company fundamentals
- tools like `get_balance_sheet`, `get_cashflow`, `get_income_statement`, and `get_insider_transactions` do not naturally fit crypto assets

I think a practical MVP could be:

1. Add `asset_type = stock | crypto`
2. In crypto mode, disable company-specific fundamentals / insider tools
[...]

PR #567, opened 2026-04-18
feat: add analysis-only crypto asset mode

CadeYu:
## Summary

This PR adds a small analysis-only crypto MVP so TradingAgents can handle symbols like `BTC-USD` and `ETH-USD` more cleanly without pretending that stock-only fundamentals are available.

The scope is intentionally limited:
- detect crypto-style symbols in the CLI
- carry `asset_type` through the graph state
- skip the fundamentals analyst in crypto mode
- make the most visible prompts more asset-oriented instead of company-only

## Why

The current workflow is still very stock/company-centric:
- the CLI assumes the same analyst set for every asset
- the state only tracks `company_of_interest`
- several prompts assume company fundamentals and company-specific context

For a first crypto MVP, it is safer to support an analysis-only path instead of trying to add full crypto-native data coverage in one step.

## Changes

- add `AssetType` and ticker-based asset detection
- classify symbols like `BTC-USD` / `ETH-USD` as `crypto`
- filter out `Fundamentals Analyst` when the selected asset is crypto
- add `asset_type` to the graph state and initial state propagation
- update the most visible analyst / researcher / trader prompts to use more neutral asset-oriented wording in crypto mode
- add unit tests for:
  - crypto symbol detection
  - analyst filtering in crypto mode
  - propagation of `asset_type` into the initial graph state

## Scope

This PR intentionally does **not**:
- add broker / execution support
- add on-chain or exchange-specific crypto data sources
- make every downstream prompt fully crypto-native
- change the stock path behavior

It is meant to be a small MVP that opens the door for future crypto-specific improvements while keeping the current stock workflow intact.

Addresses #82

## Validation

- `.venv/bin/python -m unittest tests.test_crypto_asset_mode -v`
- `.venv/bin/python -m compileall cli tradingagents tests`
- `git diff --check`


CadeYu:
Hi @Yijia-Xiao, this PR should be ready for review whenever you have time.

This is the small crypto MVP we discussed earlier in #82. I intentionally kept the scope narrow and analysis-only for the first step:

- detect crypto-style symbols such as `BTC-USD` / `ETH-USD`
- carry `asset_type` through the CLI / graph state
- skip the fundamentals analyst in crypto mode
- make the most visible prompts more asset-oriented instead of company-only

I also merged the latest `main` into this branch and resolved the conflicts, so it should now be up to date with the current base branch.

If this direction looks right, I’d be happy to follow up later with the next incremental crypto-specific improvements in separate PRs.

PR #568, opened 2026-04-18
fix(cli): resolve UnicodeEncodeError on Windows during report generation

nnishad:
- Added explicit `encoding="utf-8"` to all `open()` and `pathlib.Path.write_text()` calls in `cli/main.py`.
- Fixes application crashes on Windows when the LLM outputs non-ASCII characters (like emojis) during log writing and markdown report generation.


Yijia-Xiao:
Thanks @nnishad for the contribution! Addressed in 872b063 by switching every file I/O site to explicit encoding="utf-8". The previous PYTHONUTF8 hack had no effect because Python reads that variable only at interpreter startup.

PR #571, opened 2026-04-19
feat: persist FinancialSituationMemory to disk

### LINKED ISSUE #563: [Memory] FinancialSituationMemory is RAM-only — all lessons lost on process restart
## Summary

`FinancialSituationMemory` stores all learned lessons in two plain Python lists (`self.documents`, `self.recommendations`) and a BM25 index — with no persistence layer of any kind. This means:

1. **Every process restart wipes all memory.** The lessons the agents learned from past trades are permanently lost.
2. **Any server/API usage that creates a new `TradingAgentsGraph()` per request wipes memory on every call.** This is the natural pattern in web frameworks (FastAPI, Flask, Django), so any user hosting TradingAgents as a service gets zero benefit from the memory system.
3. **`reflect_and_remember()` is never called in the examples or CLI**, so even users running the library in long-lived scripts may not realize they need to call it — and what to pass as `returns_losses`.

## Reproducing the issue

```python
from tradingagents.graph.trading_graph import TradingAgentsGraph

# Run 1 — analysis + reflection
graph1 = TradingAgentsGraph(config=config)
state, decision = graph1.propagate("AAPL", "2024-01-15")
graph1.reflect_and_remember(0.03)  # 3% gain — lesson stored in RAM

# Run 2 — fresh process or new instance (the common server pattern)
graph2 = TradingAgentsGraph(config=config)
# graph2.bull_memory.documents == []  ← all lessons gone
state, decision = graph2.propagate("AAPL", "2024-01-22")
# Portfolio Manager's {past_memory_str} placeholder is always empty
```

## Proposed fix

Add optional persistence to `FinancialSituationMemory` via a configurable backend. The simplest path that requires no new dependencies:

**Option A — JSON file persistence** (zero dependencies):

```python
class FinancialSituationMemory:
    def __init__(self, name: str, config: dict = None):
        self.name = name
        self._persist_path = None
        if config and config.get("memory_persist_dir"):
            import pathlib
            d = pathlib.Path(config["memory_persist_dir"])
            d.mkdir(parents=True, exist_ok=True)
            self._persist_path = d / f"{name}.json"
        self.documents = []
        self.recommendations = []
        self.bm25 = None
        self._load()

    def _load(self):
        if self._persist_path and self._persist_path.exists():
            import json
            data = json.loads(self._persist_path.read_text())
            pairs = list(zip(data.get("situations", []), data.get("recommendations", [])))
            if pairs:
                self.add_situations(pairs)

    def _save(self):
        if self._persist_path:
            import json
            self._persist_path.write_text(json.dumps({
                "situations": self.documents,
                "recommendations": self.recommendations,
            }))

    def add_situations(self, situations_and_advice):
        # ... existing logic ...
        self._save()   # ← persist after every write
```

**Option B — SQLite backend** (also zero new dependencies, better for concurrent use).

**Config key to add:**

```python
DEFAULT_CONFIG = {
    ...
    "memory_persist_dir": "~/.tradingagents/memory",  # None = RAM-only (current behaviour)
    ...
}
```

## Additional request: document reflect_and_remember()

Currently `reflect_and_remember()` is not called in any example, the CLI, or the README. Users need to know:
- When to call it (after the position closes and return is known)
- What to pass as `returns_losses` (signed float — positive = signal was correct)
- That in server/API deployments, the graph instance must be kept alive between the analysis call and the reflect call, otherwise `curr_state` is lost

A short section in the README and a worked example in `main.py` would significantly improve discoverability.

Bcardo:
## Summary

Fixes #563. `FinancialSituationMemory` was RAM-only — all five memory instances (bull, bear, trader, invest judge, portfolio manager) reset to empty on every process restart.

- Add `memory_persist_dir` config key (defaults to `~/.tradingagents/memory/`)
- `FinancialSituationMemory` loads from disk on `__init__` and saves after every `add_situations()` and `clear()` call
- RAM-only mode preserved when `memory_persist_dir` is `None` or absent — no breaking change
- BM25 index is rebuilt from loaded data so retrieval works immediately after restart

## Test plan

- [ ] `pytest tests/test_memory_persistence.py -v` — 8 new tests, all passing
- [ ] `pytest tests/ -v` — 14 total, no regressions
- [ ] Manually verify `~/.tradingagents/memory/bull_memory.json` is created after calling `reflect_and_remember()`

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Claude Code local files
+CLAUDE.md
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[codz]
--- a/tests/test_memory_persistence.py
+++ b/tests/test_memory_persistence.py
@@ -1,0 +1,166 @@
+"""Tests for FinancialSituationMemory persistence (issue #563)."""
+
+import json
+import pytest
+from pathlib import Path
+
+from tradingagents.agents.utils.memory import FinancialSituationMemory
+
+
+@pytest.fixture
+def persist_dir(tmp_path):
+    return str(tmp_path / "memory")
+
+
+def make_config(persist_dir):
+    return {"memory_persist_dir": persist_dir}
+
+
+# ---------------------------------------------------------------------------
+# Persistence: data survives a fresh instance
+# ---------------------------------------------------------------------------
+
+def test_data_survives_restart(persist_dir):
+    """Documents and recommendations loaded by a new instance after save."""
+    m1 = FinancialSituationMemory("test", make_config(persist_dir))
+    m1.add_situations([("situation A", "recommendation A")])
+
+    m2 = FinancialSituationMemory("test", make_config(persist_dir))
+    assert m2.documents == ["situation A"]
+    assert m2.recommendations == ["recommendation A"]
+
+
+def test_multiple_entries_survive_restart(persist_dir):
+    """All entries are preserved across instances."""
+    m1 = FinancialSituationMemory("test", make_config(persist_dir))
+    m1.add_situations([
+        ("situation A", "rec A"),
+        ("situation B", "rec B"),
+    ])
+
+    m2 = FinancialSituationMemory("test", make_config(persist_dir))
+    assert len(m2.documents) == 2
+    assert len(m2.recommendations) == 2
+
+
+def test_bm25_index_rebuilt_on_load(persist_dir):
+    """BM25 index is functional after loading from disk."""
+    m1 = FinancialSituationMemory("test", make_config(persist_dir))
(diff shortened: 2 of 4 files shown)


hongsookim (commented):
Good implementation, but note that PR #573 by the same author is a superset of this PR (persistence + hallucination guard). Recommend merging #573 instead and closing this PR to avoid duplication.

If this PR is kept separately, the review findings are: APPROVE — JSON format, atomic writes, graceful fallback, and thorough tests are all well done.

Bcardo:
Superseded by #579 which is a complete redesign of the memory system (persistent append-only log replacing BM25). Closing to avoid review confusion.

PR #573, opened 2026-04-19
fix: persistent memory + hallucination guard for empty memory

### LINKED ISSUE #563: [Memory] FinancialSituationMemory is RAM-only — all lessons lost on process restart

### LINKED ISSUE #572: [Memory] Agents hallucinate past lessons when memory is empty
## Problem

When `past_memories` is empty (e.g. first run, or no prior `reflect_and_remember()` calls), all five agents still inject an instruction telling the LLM to *"address reflections and learn from lessons and mistakes you made in the past"* — with nothing to reference.

The LLM fulfills the instruction by fabricating plausible-sounding past experiences.

**Example from a real first-run INTU bear report:**
> "In 2022, I underestimated macro. In 2019, I overestimated AI stickiness..."

No prior runs existed. These lessons were entirely made up.

## Affected agents

- Bull Researcher
- Bear Researcher
- Trader
- Research Manager
- Portfolio Manager

## Fix

Conditionally inject the memory section only when `past_memories` is non-empty, so both the content and its surrounding instruction are absent when there is nothing to recall.

## Related

- #563 — persistent memory. The hallucination guard is the companion fix: until memory is populated via `reflect_and_remember()`, this ensures agents don't fabricate lessons from an empty corpus.

Bcardo:
## Summary

- **Persist `FinancialSituationMemory` to disk** (closes #563) — lessons learned via `reflect_and_remember()` now survive process restarts. Adds `memory_persist_dir` config key (defaults to `~/.tradingagents/memory/`); uses atomic JSON write via `tempfile` + `replace()` with graceful fallback on corruption.
- **Suppress memory injection when memory is empty** (closes #572) — all five agents previously injected an empty string into their prompts while still instructing the LLM to *"address reflections and learn from past mistakes"*, causing the LLM to hallucinate fabricated lessons on first run. Each agent now conditionally builds its memory section only when `past_memories` is non-empty.

These are companion fixes: persistence without the guard still hallucinates on first run; the guard without persistence keeps the corpus perpetually empty for CLI users.

## Changes

| File | Change |
|------|--------|
| `tradingagents/agents/utils/memory.py` | `_load()` / `_save()` persistence + import ordering fix |
| `tradingagents/default_config.py` | `memory_persist_dir` config key |
| `tradingagents/agents/researchers/bull_researcher.py` | Conditional memory section |
| `tradingagents/agents/researchers/bear_researcher.py` | Conditional memory section |
| `tradingagents/agents/trader/trader.py` | Conditional reflection clause |
| `tradingagents/agents/managers/research_manager.py` | Conditional memory section |
| `tradingagents/agents/managers/portfolio_manager.py` | Conditional lessons line |
| `tests/test_memory_persistence.py` | 10 tests for persistence (survive restart, isolation, BM25 rebuild, RAM-only, clear, corruption fallback) |
| `tests/test_hallucination_guard.py` | 10 tests for guard — empty and populated memory across all 5 agents |

## Test plan

- [x] `pytest tests/test_memory_persistence.py tests/test_hallucination_guard.py -v` — 20/20 passed
- [x] Manual run on MSFT (first run) — no fabricated past lessons in report output
- [ ] Call `ta.reflect_and_remember(returns)` after a trade, restart, run again — memory is recalled correctly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Claude Code local files
+CLAUDE.md
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[codz]
--- a/tests/test_hallucination_guard.py
+++ b/tests/test_hallucination_guard.py
@@ -1,0 +1,204 @@
+"""Tests for memory hallucination guard across all five agents.
+
+Verifies that:
+- Memory section headers are ABSENT from prompts when memory is empty.
+- Memory content IS injected when memory is populated.
+
+No LLM API calls are made — llm.invoke() is mocked.
+"""
+
+from unittest.mock import MagicMock
+
+from tradingagents.agents.researchers.bull_researcher import create_bull_researcher
+from tradingagents.agents.researchers.bear_researcher import create_bear_researcher
+from tradingagents.agents.trader.trader import create_trader
+from tradingagents.agents.managers.research_manager import create_research_manager
+from tradingagents.agents.managers.portfolio_manager import create_portfolio_manager
+
+
+# ---------------------------------------------------------------------------
+# Shared helpers
+# ---------------------------------------------------------------------------
+
+def _make_state():
+    return {
+        "investment_debate_state": {
+            "history": "",
+            "bull_history": "",
+            "bear_history": "",
+            "current_response": "",
+            "count": 0,
+        },
+        "market_report": "market data",
+        "sentiment_report": "sentiment data",
+        "news_report": "news data",
+        "fundamentals_report": "fundamentals data",
+        "company_of_interest": "AAPL",
+        "trade_date": "2024-01-15",
+        "investment_plan": "buy AAPL",
+        "trader_investment_plan": "",
+        "risk_debate_state": {
+            "history": "",
+            "aggressive_history": "",
+            "conservative_history": "",
+            "neutral_history": "",
+            "judge_decision": "",
+            "count": 0,
+            "latest_speaker": "",
+            "current_aggressive_response": "",
+            "current_conservative_response": "",
+            "current_neutral_response": "",
+        },
+    }
+
+
+def _mock_llm():
+    llm = MagicMock()
+    llm.invoke.return_value = MagicMock(content="mocked response")
+    return llm
+
+
+def _empty_memory():
+    m = MagicMock()
+    m.get_memories.return_value = []
+    return m
+
+
+def _populated_memory(lesson="Past lesson: watch macro risk"):
+    m = MagicMock()
+    m.get_memories.return_value = [{"recommendation": lesson, "similarity_score": 0.9}]
+    return m
+
+
+# ---------------------------------------------------------------------------
+# Bull Researcher
+# ---------------------------------------------------------------------------
+
+def test_bull_omits_memory_section_when_empty():
+    """No reflections header or instruction when memory is empty."""
+    llm = _mock_llm()
+    node = create_bull_researcher(llm, _empty_memory())
+    node(_make_state())
+
+    prompt = llm.invoke.call_args[0][0]
+    assert "Reflections from similar situations" not in prompt
+    assert "address reflections" not in prompt.lower()
+    assert "learn from lessons and mistakes" not in prompt.lower()
+
+
+def test_bull_includes_memory_section_when_populated():
+    """Lesson text appears in prompt when memory is populated."""
+    llm = _mock_llm()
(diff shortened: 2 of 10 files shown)


hongsookim (approved):
Solid PR that addresses both #572 (hallucination) and #563 (persistence). The two-part approach is architecturally sound — persistence without the guard still hallucinates on first run, and vice versa.

**Hallucination guard:** Correctly omits both memory data AND instructions when memory is empty, across all 5 agents. This is the right approach.

**Persistence:** JSON + atomic writes via tempfile + replace(), graceful fallback on corruption, backward-compatible RAM-only mode.

**Tests:** Thorough — 10 persistence + 10 hallucination guard tests covering all agents in both empty and populated states.

**One recommended fix (minor):**
- `_save()`: move `mkdir` inside the `try` block and add temp file cleanup if `replace()` fails, to prevent temp-file leaks (as Gemini also suggested)

**Optional improvements:**
- Switch from `"".join(rec + "\n\n")` to `"\n\n".join(rec)` to avoid trailing separators
- Refactor bull/bear ternary expressions into explicit `if/else` blocks for consistency with the other 3 agents

Approving — the _save() fix can be addressed in a follow-up if needed.

Bcardo:
Superseded by #579 which is a complete redesign of the memory system (persistent append-only log replacing BM25). Closing to avoid review confusion.

PR #577, opened 2026-04-19
feat: add MiniMax OpenAI-compatible provider support

Jubeing:
## Summary

Adds MiniMax as a new LLM provider using the OpenAI-compatible endpoint.

### Changes

- **`tradingagents/llm_clients/factory.py`**: Added `minimax` to `_OPENAI_COMPATIBLE` tuple
- **`tradingagents/llm_clients/openai_client.py`**: Added minimax entry to `_PROVIDER_CONFIG`:
  - Base URL: `https://api.minimaxi.com/v1`
  - Auth env var: `MINIMAX_CN_API_KEY`

### Usage

```python
config = DEFAULT_CONFIG.copy()
config["llm_provider"] = "minimax"
config["deep_think_llm"] = "MiniMax-M2.7"
config["quick_think_llm"] = "MiniMax-M2.7"

ta = TradingAgentsGraph(debug=True, config=config)
decision = ta.propagate("NVDA", "2024-05-10")
```

### Environment

Set `MINIMAX_CN_API_KEY` in your `.env` file. Get your key at https://platform.minimaxi.com

---

Tested with MiniMax-M2.7 model via OpenAI-compatible chat completions endpoint.

### DIFF
--- a/tradingagents/llm_clients/factory.py
+++ b/tradingagents/llm_clients/factory.py
@@ -9,6 +9,7 @@ from .azure_client import AzureOpenAIClient
 # Providers that use the OpenAI-compatible chat completions API
 _OPENAI_COMPATIBLE = (
     "openai", "xai", "deepseek", "qwen", "glm", "ollama", "openrouter",
+    "minimax",
 )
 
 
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @Jubeing! Same provider landed in 19d22b5, scaled to cover both regional endpoints (global .io + China .com) rather than CN-only.

PR #582, opened 2026-04-20
feat: spend limits with automatic abort (#524)

### LINKED ISSUE #524: Spend limits and scope enforcement for trading agents — delegation chains with audit trail
A trading agent that can place unlimited orders with no scope constraints is a single prompt injection away from draining an account. The current architecture trusts the LLM's judgment as the only gate between "analyze AAPL" and "market sell entire portfolio."

Three failure modes that delegation scoping prevents:

**1. Spend limit enforcement.** A research agent shouldn't be able to execute trades. A trading agent should have a per-trade and daily cap. These constraints need to be cryptographically enforced, not just checked in application logic that the agent's own code could bypass.

**2. Asset scope.** An agent authorized to trade equities shouldn't be able to touch options or crypto. Scope narrowing means the delegation specifies exactly which instruments and actions are permitted.

**3. Audit trail for compliance.** When the SEC asks "why did your agent execute this trade," you need more than log files. You need a signed receipt chain proving the agent had authority, what parameters were checked, and who delegated that authority.

```typescript
import { createDelegation, governAction } from 'agent-passport-system'

const tradingDelegation = createDelegation({
  delegatedTo: agentKey,
  delegatedBy: portfolioManagerKey,
  scope: ['trade:equity:buy', 'trade:equity:sell'],  // no options, no crypto
  spendLimit: 100000,   // $1,000 max per day
  expiresAt: new Date(Date.now() + 8 * 3600_000),  // market hours only
  maxDepth: 0  // cannot sub-delegate to other agents
})

// Every trade goes through enforcement
const result = await governAction(
  { type: 'trade:equity:buy', ticker: 'AAPL', amount: 15000 },
  async (action) => broker.executeTrade(action),
  { passport, delegation: tradingDelegation, privateKey: agentKey }
)
// result.receipt — signed proof: who authorized, what scope, what happened
// result.spentAmount updated — approaching limit triggers circuit breaker
```

The delegation chain is the compliance artifact. Each receipt in the chain is Ed25519-signed, timestamped, and includes the full scope context at evaluation time. An auditor can reconstruct exactly what authority existed at any point.

`npm install agent-passport-system` (v1.36.2, Apache-2.0). Gateway at gateway.aeoess.com for hosted enforcement with real-time dashboards.

The multi-agent structure in TradingAgents maps naturally to delegation hierarchies: portfolio manager delegates to sector analysts, analysts delegate to execution agents, each with progressively narrower scope. Authority can only decrease at each level.

claytonbrown:
Add --max-cost flag (USD) to cap LLM spend per analysis run. When the budget is exceeded mid-graph, the analysis aborts gracefully with partial results saved.

Components:
- SpendTracker: LangChain callback handler tracking cumulative token usage/cost
- BudgetExceededError: raised when budget exceeded, caught by graph runner
- Delegation chain audit trail: which agent called which tool/LLM, cost per call
- CLI: real-time cost display, --max-cost flag
- Model pricing table for cost estimation

Tests verify abort triggers at budget and partial results are preserved.

Closes #524

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -253,7 +253,7 @@ def format_tokens(n):
     return str(n)
 
 
-def update_display(layout, spinner_text=None, stats_handler=None, start_time=None):
+def update_display(layout, spinner_text=None, stats_handler=None, start_time=None, spend_tracker=None):
     # Header with welcome message
     layout["header"].update(
         Panel(
@@ -444,6 +444,15 @@ def update_display(layout, spinner_text=None, stats_handler=None, start_time=Non
         else:
             tokens_str = "Tokens: --"
         stats_parts.append(tokens_str)
+
+    # Cost display from spend tracker
+    if spend_tracker:
+        cost_str = f"${spend_tracker.total_cost_usd:.4f}"
+        if spend_tracker.max_cost is not None:
+            cost_str += f"/${spend_tracker.max_cost:.2f}"
+        if spend_tracker.budget_exceeded:
+            cost_str = f"[red]{cost_str} OVER BUDGET[/red]"
+        stats_parts.append(f"Cost: {cost_str}")
 
     stats_parts.append(f"Reports: {reports_completed}/{reports_total}")
 
@@ -926,7 +935,7 @@ def format_tool_args(args, max_length=80) -> str:
         return result[:max_length - 3] + "..."
     return result
 
-def run_analysis():
+def run_analysis(max_cost: float | None = None):
     # First get all user selections
     selections = get_user_selections()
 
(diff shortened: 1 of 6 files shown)


Yijia-Xiao:
Thanks @claytonbrown for the thoughtful design! Closing without merge: cumulative token usage is already tracked via cli/stats_handler.py, and a hardcoded pricing table would create stale-data risk (provider prices change frequently; cached tokens, reasoning surcharges, and Anthropic prompt caching change effective rates). Without the pricing layer the budget abort lacks a meaningful trigger. The callback handler + abort + partial-save shape you built is the right starting point if we revisit this with user-supplied rates.

PR #583, opened 2026-04-20
feat: Pydantic schema validation for agent outputs (#434)

claytonbrown:
Add structured output validation to all agent nodes. Malformed LLM responses (missing rating, invalid format) are caught and retried with error feedback.

Components:
- schemas.py: Pydantic models for AnalystReport, TraderDecision, RiskAssessment, PortfolioDecision
- output_parser.py: validates LLM response against schema, retries up to 2x with error context
- Wired into all 13 agent nodes (4 analysts, 2 researchers, 3 risk, trader, 2 managers, portfolio)

Tests verify malformed output triggers retry and produces valid schema.

Closes #434

### DIFF
--- a/tests/test_output_parser.py
+++ b/tests/test_output_parser.py
@@ -1,0 +1,144 @@
+"""Tests for structured output parsing with retry on malformed output."""
+
+import unittest
+
+from pydantic import ValidationError
+
+from tradingagents.agents.output_parser import StructuredOutputParser, validate_agent_output
+from tradingagents.agents.schemas import (
+    AnalystReport,
+    PortfolioDecision,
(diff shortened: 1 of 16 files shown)


Yijia-Xiao:
Thanks @claytonbrown for the careful contribution! Structured output has now landed across the three decision-making agents in 0fda245 and bba1477, with a different design choice: each agent's primary call uses llm.with_structured_output(Schema) (canonical, single LLM call per agent) rather than the JSON-coercion-via-retry pattern. Analyst, researcher, and risk-debater agents stay prose-first since their narrative is read by users and downstream LLM agents. Closing #434 alongside.

PR #594, opened 2026-04-22
feat: LangGraph checkpoint resume for crash recovery

claytonbrown:
Add SqliteSaver-based checkpointing so crashed analyses resume from the last successful graph node instead of restarting from scratch.

- checkpointer.py: get_checkpointer(), thread_id(), has/clear_checkpoint()
- --checkpoint flag (default: off for backward compatibility)
- --clear-checkpoints flag to force fresh start
- Per-ticker SQLite DBs for parallel worker safety
- Logs 'Resuming from step N' vs 'Starting fresh'
- Clears checkpoint on successful completion (no stale state)
- Tests: crash resume + different date starts fresh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -217,3 +217,6 @@ __marimo__/
 
 # Cache
 **/data_cache/
+.kiro/
+ralph-kiro.sh
+progress.txt
--- a/README.md
+++ b/README.md
@@ -219,6 +219,50 @@ print(decision)
 
 See `tradingagents/default_config.py` for all configuration options.
 
+## Checkpoint / Resume
+
+TradingAgents can save progress after each graph node so that a crashed or interrupted run resumes from the last successful step instead of starting over.
+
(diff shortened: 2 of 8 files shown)


Yijia-Xiao:
Thanks @claytonbrown for the careful design and the resume tests! Applied in 4cbd4b0. Your SqliteSaver-per-ticker shape, the deterministic thread_id from ticker+date so the same date resumes while a different date starts fresh, and the clear-on-success pattern all carried through. The integration was layered on top of our recent memory-log redesign: the graph_setup now returns the uncompiled workflow so propagate() can recompile with a saver when --checkpoint is enabled, and the existing memory-log lifecycle (resolve pending entries at start, store decision at end) is preserved end to end. Documentation is in README under 'Checkpoint / Resume', and langgraph-checkpoint-sqlite is added as a project dependency.

PR #596, opened 2026-04-23
fix: suppress memory prompt when past_memories is empty (fixes #572)

### LINKED ISSUE #572: [Memory] Agents hallucinate past lessons when memory is empty

octo-patch:
Fixes #572

## Problem

When `past_memories` is empty (first run, or before any `reflect_and_remember()` calls), all five agents still inject an instruction telling the LLM to *"address reflections and learn from lessons and mistakes you made in the past"* — with nothing to reference.

The LLM fulfills this instruction by fabricating plausible-sounding past experiences, as described in the issue:
> "In 2022, I underestimated macro. In 2019, I overestimated AI stickiness..."

## Solution

Conditionally include both the memory content **and** the surrounding instruction only when `past_memories` is non-empty. When memory is empty, neither the content nor the directive to reference it is injected, so agents have no prompt to fabricate from.

## Affected agents

- Bull Researcher
- Bear Researcher
- Research Manager
- Portfolio Manager
- Trader

## Testing

On first run (empty memory), none of the five agents will receive the memory instruction or placeholder content. On subsequent runs after `reflect_and_remember()` has been called, the memory section and instruction are injected as before.

### DIFF
--- a/tradingagents/agents/managers/portfolio_manager.py
+++ b/tradingagents/agents/managers/portfolio_manager.py
@@ -22,6 +22,12 @@ def create_portfolio_manager(llm, memory):
         for i, rec in enumerate(past_memories, 1):
             past_memory_str += rec["recommendation"] + "\n\n"
 
+        lessons_line = (
+            f"- Lessons from past decisions: **{past_memory_str}**\n"
+            if past_memories
+            else ""
+        )
+
         prompt = f"""As the Portfolio Manager, synthesize the risk analysts' debate and deliver the final trading decision.
 
 {instrument_context}
@@ -38,8 +44,7 @@ def create_portfolio_manager(llm, memory):
 **Context:**
 - Research Manager's investment plan: **{research_plan}**
 - Trader's transaction proposal: **{trader_plan}**
-- Lessons from past decisions: **{past_memory_str}**
-
+{lessons_line}
 **Required Output Structure:**
 1. **Rating**: State one of Buy / Overweight / Hold / Underweight / Sell.
 2. **Executive Summary**: A concise action plan covering entry strategy, position sizing, key risk levels, and time horizon.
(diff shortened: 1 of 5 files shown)


octo-patch:
Thanks for the review! Pushed 25cc0e1 with both rounds of feedback applied:

1. **Stripped trailing newlines** on `past_memory_str` in all five touched files (portfolio_manager, research_manager, bear/bull researchers, trader) so the `"\n\n"` separators between recommendations no longer leak into the prompt boundaries.
2. **Made the Investment Thesis instruction conditional** in `portfolio_manager.py` — the "and past reflections" phrasing is now only included when `past_memories` is non-empty, matching the existing memory-section gate and preventing the hallucination vector flagged in the review.

Yijia-Xiao:
Thanks @octo-patch for the careful and well-targeted contribution! Applied in 8e7654f. The directive and the reference line are both removed from the prompt when memory is empty, so agents are no longer asked about past lessons that do not exist. Verified across all five agents that consume memory.

PR #600, opened 2026-04-26
Fix deepseek-v4 reasoning model compatibility

Bronny-62:
## Summary

- Add DeepSeek V4 Pro and V4 Flash to the shared model catalog and model validation.
- Preserve DeepSeek reasoning_content across chat completion turns so thinking-mode models can continue after tool calls.
- Skip unsupported structured-output tool_choice binding for deepseek-reasoner and use the existing free-text fallback path.
- Ignore local macOS .DS_Store files and generated CLI reports.

## Tests

- .venv/bin/python -m pytest tests/test_openai_client_config.py tests/test_model_validation.py

## Manual verification

- Verified DeepSeek reasoner tool-call loop no longer fails with reasoning_content 400.
- Verified CLI analysis can complete and output a Portfolio Management Decision.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -217,3 +217,9 @@ __marimo__/
 
 # Cache
 **/data_cache/
+
+# macOS
+.DS_Store
+
+# Locally exported CLI reports
+reports/
--- a/tests/test_model_validation.py
+++ b/tests/test_model_validation.py
@@ -53,3 +53,11 @@ class ModelValidationTests(unittest.TestCase):
                     client.get_llm()
 
                 self.assertEqual(caught, [])
+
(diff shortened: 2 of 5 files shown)


Yijia-Xiao:
Thanks @Bronny-62 for the careful design and tests! The reasoning_content round-trip from this PR landed in 7e9e7b8, refactored into a DeepSeekChatOpenAI subclass (extending the existing NormalizedChatOpenAI) so the provider-specific behaviour is cleanly isolated from the general client. The non-list-input case the gemini bot flagged is also handled via a small _input_to_messages helper. Thanks again.

PR #611, opened 2026-04-28
fix: preserve reasoning_content for DeepSeek V4 thinking mode

### LINKED ISSUE #599: Bug: DeepSeek V4 reasoning_content error when using thinking mode
When using DeepSeek V4 with thinking mode enabled, the following error occurs:

```
 BadRequestError: Error code: 400 - {'error': {'message': 'The reasoning_content in the thinking mode must be passed back to the API.', 'type': 'invalid_request_error', 'param': None, 'code':
 'invalid_request_error'}}

```

###  Steps to reproduce

 1. Configure TradingAgents to use DeepSeek V4 model
 2. Enable thinking mode (e.g., reasoning_effort: "high")
 3. Run any agent (e.g., Market Analyst)

### Expected behavior

 Should work without errors
### Actual behavior
```
 BadRequestError with 400 status code
```
### Root cause (likely)

 DeepSeek V4's thinking mode requires reasoning_content to be passed back in subsequent API calls. TradingAgents' langchain wrapper doesn't preserve this field in message history.
 Environment
 - TradingAgents version: (latest)
 - Model: deepseek-chat (V4)
 - Provider: deepseek

octo-patch:
Fixes #599

## Problem

When using DeepSeek V4 models (`deepseek-v4-flash`, `deepseek-v4-pro`) with thinking mode enabled (e.g. `reasoning_effort: "high"`), subsequent API calls fail with:

```
BadRequestError: 400 - The `reasoning_content` in the thinking mode must be passed back to the API.
```

DeepSeek's API requires that when a response includes `reasoning_content`, that field must be echoed back in the assistant message on the next turn. LangChain's current implementation drops this field in both directions:

- **Receiving**: `_convert_dict_to_message` does not capture `reasoning_content` into `AIMessage.additional_kwargs`
- **Sending**: `_convert_message_to_dict` does not include `reasoning_content` when serializing assistant messages

## Solution

Override two methods in `NormalizedChatOpenAI`:

1. **`_create_chat_result`** — after the parent converts the raw response, extracts `reasoning_content` from each choice's message dict and stores it in `AIMessage.additional_kwargs["reasoning_content"]`.

2. **`_get_request_payload`** — after the parent builds the request payload, injects `reasoning_content` back into any assistant message dict where it is present in `additional_kwargs` but missing from the converted dict.

This is a minimal, targeted shim that does not affect other providers. The default DeepSeek V3 model (no thinking mode) is unaffected since `reasoning_content` is absent from its responses.

## Testing

- Imports verified clean
- Existing model validation tests pass (`tests/test_model_validation.py`)
- Logic manually traced against langchain-openai 0.3.35 source for `_create_chat_result` and `_get_request_payload`

### DIFF
--- a/tradingagents/agents/managers/portfolio_manager.py
+++ b/tradingagents/agents/managers/portfolio_manager.py
@@ -22,6 +22,12 @@ def create_portfolio_manager(llm, memory):
         for i, rec in enumerate(past_memories, 1):
             past_memory_str += rec["recommendation"] + "\n\n"
 
+        lessons_line = (
+            f"- Lessons from past decisions: **{past_memory_str}**\n"
+            if past_memories
+            else ""
+        )
+
         prompt = f"""As the Portfolio Manager, synthesize the risk analysts' debate and deliver the final trading decision.
 
 {instrument_context}
@@ -38,8 +44,7 @@ def create_portfolio_manager(llm, memory):
 **Context:**
 - Research Manager's investment plan: **{research_plan}**
 - Trader's transaction proposal: **{trader_plan}**
-- Lessons from past decisions: **{past_memory_str}**
-
+{lessons_line}
 **Required Output Structure:**
 1. **Rating**: State one of Buy / Overweight / Hold / Underweight / Sell.
 2. **Executive Summary**: A concise action plan covering entry strategy, position sizing, key risk levels, and time horizon.
--- a/tradingagents/agents/managers/portfolio_manager.py
+++ b/tradingagents/agents/managers/portfolio_manager.py
@@ -23,9 +23,14 @@ def create_portfolio_manager(llm, memory):
             past_memory_str += rec["recommendation"] + "\n\n"
 
         lessons_line = (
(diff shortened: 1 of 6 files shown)


djmills1221-cyber (approved):

octo-patch:
Addressed in 67582de — switched to `if reasoning is not None and isinstance(gen.message, AIMessage):` so an empty `reasoning_content` is still echoed back per DeepSeek's API requirement.

The unused `i` loops in `portfolio_manager`, `research_manager`, `bear_researcher`, `bull_researcher`, and `trader` are pre-existing — outside the scope of this fix, leaving them for a separate cleanup.

Yijia-Xiao:
Thanks @octo-patch for the careful contribution! Closing in favour of the design that landed in 7e9e7b8. The reasoning_content propagation idea is the same; we used the alternative implementation because the agent-file changes in this PR overlap with the v0.2.4 memory redesign (the past_memories prompt branches were removed in #579/ebd2e12 and reintroducing them here would regress that work). Appreciate the contribution.

PR #622, opened 2026-04-30
feat(polygon): Polygon as default vendor + Anthropic thinking/structured-output bug fix

justinpreston:
## Summary

Replaces yfinance with Polygon as the default vendor for fundamentals, bars, indicators, and news. yfinance + alpha_vantage remain as automatic fallbacks via the existing `route_to_vendor` chain. Also bundles the Anthropic extended-thinking + structured-output incompatibility fix surfaced in the 3-way 2024-05-10 run.

### Bug 1: Path D NVDA market cap was $224.6B (10x low)
The yfinance Path D derivation multiplied split-adjusted historical price by pre-split share count, producing a 10x-low market cap on NVDA at 2024-05-10. Polygon's `/v3/reference/tickers?date=X` endpoint returns the as-of market_cap pre-computed correctly, eliminating the bug class.

**Verified:** `route_to_vendor('get_fundamentals', 'NVDA', '2024-05-10')` now returns Market Cap **$2.25T** (matches reality).

### Bug 2: Anthropic thinking + structured output rejected
`claude-opus-4.7-xhigh` (and other thinking-enabled Claude variants) fail with `400 - Thinking may not be enabled when tool_choice forces tool use` when LangChain's `with_structured_output` imposes `tool_choice='any'`. Affected the Research Manager during persona-aligned runs, silently degrading to free-text plans.

**Fix:** `bind_structured` detects Claude models with effort suffixes (`-xhigh`, `-high`, `-medium`, `-low`, `-minimal`) and builds a non-thinking *twin* via `model_copy` for structured calls only. The original thinking-enabled LLM remains in use for free-text and other paths.

## Polygon-specific design decisions

- **PIT visibility filter** (`_is_pit_visible`): `filing_date.lt` does not exclude null filing_date rows, so we combine with `period_of_report_date.lt` AND a Python-side filter using a 90-day filing-lag fallback when filing_date is null.
- **Strict TTM** (`_ttm_sum`): returns None if any of 4 most-recent quarters is missing the concept value, preventing partial-window misleading aggregates.
- **Cache key versioned by vendor**: `{symbol}-{Polygon|YFin}-data-{start}-{end}.csv` so switching vendors invalidates stale files.
- **Insider transactions**: not on Stocks Starter plan; `get_insider_transactions` raises `PolygonError` so the router falls through to yfinance.

## Tests

143 passing (+ 38 subtests).

- `tests/test_polygon_finance.py` — 21 tests: PIT rule, TTM strictness, statement-to-CSV, integration with mocked endpoints, vendor fallback chain.
- `tests/test_anthropic_thinking_structured.py` — 18 tests: suffix stripping, twin construction (ChatOpenAI + ChatAnthropic shapes), clone-failure fallback, bind_structured integration.

### DIFF
--- a/tests/test_anthropic_thinking_structured.py
+++ b/tests/test_anthropic_thinking_structured.py
@@ -1,0 +1,184 @@
+"""Regression tests for the Anthropic-thinking + structured-output fix.
+
+When Claude is served through GitHub Copilot (a ``ChatOpenAI``-shaped
+client pointed at ``api.githubcopilot.com``), the model id encodes the
+extended-thinking effort level via a suffix:
+``claude-opus-4.7-xhigh``. Any structured-output call against that model
+fails with::
+
+    400 - Thinking may not be enabled when tool_choice forces tool use.
+
+``bind_structured`` is supposed to detect this combination and build a
+*non-thinking twin* (same model id with the suffix stripped) that can
+accept the forced ``tool_choice`` LangChain emits for structured output.
+The original thinking-enabled LLM stays in use everywhere else.
+
+These tests stub the LLM so they don't hit the network — what we care
+about is that the routing logic picks the right twin and falls back to
+the original LLM when no remediation is possible.
+"""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock
+
+import pytest
+from pydantic import BaseModel
+
+from tradingagents.agents.utils import structured as struct_mod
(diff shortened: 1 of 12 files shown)


justinpreston:
Auto-closed: opened against upstream by mistake. Reopening on the fork (justinpreston/TradingAgents).

PR #645, opened 2026-05-01
fix: add yf_retry to _fetch_returns, per-entry error handling, remove dead entitlement code

angziii:
## Summary

Three small fixes that improve error resilience and remove dead code:

- **Add `yf_retry` to `_fetch_returns` yfinance calls.** The rest of the dataflows layer uses `yf_retry` with exponential backoff to handle Yahoo Finance HTTP 429 rate limits, but `_fetch_returns` called `yf.Ticker(...).history()` directly. A rate-limit response during memory log resolution would crash the entire `propagate()` call.

- **Add per-entry error handling in `_resolve_pending_entries`.** The `reflect_on_final_decision` call (which invokes an LLM) was outside any try/except, so a single failing reflection discarded the entire batch of pending updates. Now a failed reflection logs a warning and skips to the next entry, matching the existing resilience pattern used for `_fetch_returns`.

- **Remove dead `_current_entitlement` global lookup in `alpha_vantage_common.py`.** `_make_api_request` called `globals().get('_current_entitlement')` but `_current_entitlement` is never defined anywhere in the project. Removed the dead code and simplified the entitlement logic.

## Test plan

- [x] Existing tests pass (no regressions)
- [ ] Manual: run `tradingagents analyze` with yfinance data vendor to confirm `propagate()` still works
- [ ] Manual: trigger a rate limit during memory log resolution to confirm retry kicks in

### DIFF
--- a/tradingagents/dataflows/alpha_vantage_common.py
+++ b/tradingagents/dataflows/alpha_vantage_common.py
@@ -53,10 +53,9 @@ def _make_api_request(function_name: str, params: dict) -> dict | str:
         "source": "trading_agents",
     })
     
-    # Handle entitlement parameter if present in params or global variable
-    current_entitlement = globals().get('_current_entitlement')
-    entitlement = api_params.get("entitlement") or current_entitlement
-    
+    # Handle entitlement parameter if present in params
+    entitlement = api_params.get("entitlement")
+
     if entitlement:
         api_params["entitlement"] = entitlement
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @angziii. Fixed in 60dcf64: reflection is guarded per entry, so a provider error leaves that decision pending instead of stopping the run that follows. The return fetch was already guarded.

PR #648, opened 2026-05-01
Allow OLLAMA_BASE_URL env var to override hardcoded localhost

mstxus:
The ollama provider was hardcoded to http://localhost:11434/v1, which prevented running TradingAgents on a remote machine while serving Ollama on a different host (e.g., a separate VM, a Mac with Apple Silicon GPU acceleration, or any non-localhost setup).

This change reads OLLAMA_BASE_URL from the environment when set, falling back to the existing localhost default. No behavior change for existing users; opt-in via env var for remote setups.

Tested with Ollama running on a remote Mac while TradingAgents executed in a Linux VM.


Yijia-Xiao:
Thanks @mstxus — shipped in f10daa2. Your suggestion (read `OLLAMA_BASE_URL` from the environment, fall back to localhost) is the right shape; we adopted it across both the programmatic client (`_resolve_provider_base_url` in `openai_client.py`) and the CLI dropdown (`select_llm_provider` in `cli/utils.py`), so it works for either entry point. The CLI also now prints the resolved endpoint after Ollama is selected, marks env-origin, and warns on common malformed inputs (missing scheme, non-default port on remote hosts). See `.env.example` for the canonical setup.

PR #649, opened 2026-05-01
Fix CLI display of literal-looking message text

mrbob-git:
What I changed
- Keep string message content as text in the CLI display, even when it looks like a Python falsy literal such as `0` or `False`.
- Still treats real empty containers and blank strings as empty.
- Adds focused unit coverage for the display helper.

How to verify
- `python -m pytest`

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -855,20 +855,12 @@ def extract_content_string(content):
     """Extract string content from various message formats.
     Returns None if no meaningful text content is found.
     """
-    import ast
-
     def is_empty(val):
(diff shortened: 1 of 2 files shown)


mrbob-git:
Picked up the test coverage suggestion and kept the production change unchanged. Local checks are still passing with `python -m pytest`.

Yijia-Xiao:
Thanks @mrbob-git. Fixed in 6398951: a message is judged empty only when nothing was written, so text like "0" or "None" is shown.

PR #659, opened 2026-05-02
Enhance smoke test with report and timing features

krataratha:
Added features for JSON report export, execution timing, and quiet mode.


Tauric-Research:

PR #666, opened 2026-05-02
feat(cli): scrollable Textual TUI for live analyze view

sgundava:
Replace the fixed-height Rich Live layout with a Textual app that renders the same panes (header / progress / messages / current report / footer stats) but routes the messages and report through scrollable containers, so long reports and earlier tool calls are no longer truncated. Mouse wheel, arrow keys, g/G for top/bottom, and Tab/Shift-Tab to switch panes.

The graph stream loop is blocking, so it runs in a Textual worker thread (@work(thread=True)) and the UI refreshes on a 0.25s interval — same cadence as the classic refresh_per_second=4. The classic Rich Live renderer is preserved verbatim behind --classic (also TRADINGAGENTS_CLASSIC_TUI=1) for one release so any scripts depending on the previous output have a fallback.

Refactor: chunk → buffer mapping (~95 lines that lived inline in the stream loop) is extracted to handle_stream_chunk(buffer, chunk), shared by both renderers. update_research_team_status now takes a buffer arg instead of relying on the module global.

- New cli/tui.py: TradingApp + pure Rich-renderable builders (build_progress_table, build_messages_table, build_stats_text) exercised by a headless app.run_test smoke check.
- New cli/tui.tcss: layout + colours mirroring the classic panels (cyan progress, blue messages, green report).
- pyproject.toml: add textual>=0.80, ship *.tcss with the package.
- .gitignore: ignore .DS_Store on macOS.
- CHANGELOG: entry under [Unreleased].

All 108 existing tests still pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -221,3 +221,6 @@ __marimo__/
 # Enterprise env file (secrets) and generated run reports
 .env.enterprise
 reports/
+
+# macOS
+.DS_Store
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,28 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 Breaking changes within the 0.x line are called out explicitly.
 
+## [Unreleased]
+
+### Added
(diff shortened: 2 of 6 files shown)


Tauric-Research:

PR #673, opened 2026-05-03
feat: add investment_horizon configuration parameter

### LINKED ISSUE #670: Test

ankitadubey323:
Closes #670

## Changes
- Added `investment_horizon` config parameter in `default_config.py`
- Added `INVESTMENT_HORIZONS` constant for supported horizon types
- Added CLI prompt for horizon selection (Step 9) in `cli/main.py`
- Added `ask_investment_horizon()` function in `cli/utils.py`
- Added `get_horizon_instruction()` helper in `agents/utils/agent_utils.py`
- Updated all 4 analyst prompts to adapt analysis based on investment horizon

## Testing
- Default `medium_term` is backward compatible — no breaking changes
- Tested horizon guidance with `1_day` and `5_years_plus` values

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,3 +269,5 @@ Please reference our work if you find *TradingAgents* provides you with some hel
       url={https://arxiv.org/abs/2412.20138}, 
 }
 ```
+## Beginner Note
+This project is a multi-agent trading system using AI models.
--- a/cli/main.py
+++ b/cli/main.py
@@ -565,7 +565,7 @@ def get_user_selections():
     selected_shallow_thinker = select_shallow_thinking_agent(selected_llm_provider)
     selected_deep_thinker = select_deep_thinking_agent(selected_llm_provider)
 
-
+    # Step 9: Investment Horizon
     console.print(
         create_question_box(
             "Step 9: Investment Horizon",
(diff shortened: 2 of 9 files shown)


Ghraven:
Hey @ankitadubey323, interesting addition — horizon-aware prompt injection is a sensible way to steer agent behavior without overhauling the whole pipeline.

A couple of observations from the diff:

1. **`medium_term` not in `INVESTMENT_HORIZONS`** — the constant in `default_config.py` defines six keys (`1_day` through `5_years_plus`) but the default value is `"medium_term"`, which is not among them. This means a fresh run with the default config hits a `.get()` miss and falls through to the `"medium_term"` branch inside `get_horizon_instruction()` — but it is not mapped in the dict either, so the function would return an empty string silently. Worth either adding `"medium_term"` to `INVESTMENT_HORIZONS` or changing the default to one of the defined keys like `"1_month"`.

2. **Step numbering in CLI** — the new prompt is labelled "Step 9" but it is inserted before the existing provider-specific thinking config which was presumably Step 8. Might cause confusion for users following docs or tutorials with step references.

Otherwise the `get_horizon_instruction()` approach of returning a guidance string and concatenating it to existing prompts is clean — no invasive changes to the agent structure.

ankitadubey323:
Thanks for the review @Ghraven!
Fixed both issues:
1. Added `medium_term` to `INVESTMENT_HORIZONS` in `default_config.py`
2. Moved Investment Horizon to Step 8, before provider-specific thinking config. Renumbered subsequent steps accordingly.

Ghraven:
Great work @ankitadubey323 — both fixes look correct. Adding `medium_term` to `INVESTMENT_HORIZONS` closes the silent-failure gap, and the step renumbering is clean. LGTM from my side 👍

PR #689, opened 2026-05-03
fix: raise NotImplementedError for all DeepSeek models in structured output

### LINKED ISSUE #678: AI Manager structured-output invocation failed
when i using DeepSeek V4 Flash and DeepSeek V4 Pro, the cli give me the info:
`
? Select Your [Quick-Thinking LLM Engine]: DeepSeek V4 Flash - Latest V4 fast model
? Select Your [Deep-Thinking LLM Engine]: DeepSeek V4 Pro - Latest V4 flagship model
Research Manager: structured-output invocation failed (Error code: 400 - {'error': {'message': 'deepseek-reasoner does
not support this tool_choice', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_request_error'}});
retrying once as free text
Trader: structured-output invocation failed (Error code: 400 - {'error': {'message': 'deepseek-reasoner does not support
this tool_choice', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_request_error'}}); retrying once as
free text
Portfolio Manager: structured-output invocation failed (Error code: 400 - {'error': {'message': 'deepseek-reasoner does
not support this tool_choice', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_request_error'}});
retrying once as free text
`

HrushiYadav:
Closes #678.

All DeepSeek models reject the `tool_choice` parameter with HTTP 400, not just `deepseek-reasoner`. Previously other variants like `deepseek-chat` would bind successfully but fail at invocation, wasting an API call and logging an error before falling back to free-text. Changed `with_structured_output()` to raise `NotImplementedError` for all DeepSeek models so `bind_structured()` catches it cleanly upfront. Updated tests to cover all three variants.

### DIFF
--- a/tests/test_deepseek_reasoning.py
+++ b/tests/test_deepseek_reasoning.py
@@ -5,9 +5,9 @@ Two pieces verified:
 1. ``reasoning_content`` is captured on receive into the AIMessage's
    ``additional_kwargs`` and re-attached on send so DeepSeek's API
    sees the same value across turns.
-2. ``with_structured_output`` raises NotImplementedError for
-   ``deepseek-reasoner`` so the agent factories' free-text fallback
-   handles the request instead of failing at runtime.
+2. ``with_structured_output`` raises NotImplementedError for all
+   DeepSeek models so the agent factories' free-text fallback handles
+   the request instead of failing at runtime with HTTP 400.
 """
 
 import os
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @HrushiYadav for surfacing this! Closed in favor of a different approach in 22bb91b: instead of raising NotImplementedError for all DeepSeek models (which would lose structured output for deepseek-chat where tool_choice works, and force every thinking-mode user into free-text fallback), we route through a declarative capability table that suppresses just the tool_choice parameter for models that reject it. The schema is still bound as a tool — exactly matching DeepSeek's official tool-calling examples — so V4 Pro / V4 Flash / reasoner all keep working structured output.

Issue #627, opened 2026-04-30
PermissionError: [Errno 13] Permission denied: '/home/appuser/.tradingagents/cache'


suyashbansal:
I installed the project via Docker and tested it using Gemini as the provider. At the step where it asks to select the thinking mode, I chose the recommended option (Enable Thinking), but it ran into an error.

Am I missing any setup steps here?

```shell
? Select Thinking Mode: Enable Thinking (recommended)
╭────────────────────────────────────────────────────────── Traceback (most recent call last) ──────────────────────────────────────────────────────────╮
│ /opt/venv/lib/python3.12/site-packages/cli/main.py:1217 in analyze                                                                                    │
│                                                                                                                                                       │
│   1214 │   │   from tradingagents.graph.checkpointer import clear_all_checkpoints                                                                     │
│   1215 │   │   n = clear_all_checkpoints(DEFAULT_CONFIG["data_cache_dir"])                                                                            │
│   1216 │   │   console.print(f"[yellow]Cleared {n} checkpoint(s).[/yellow]")                                                                          │
│ ❱ 1217 │   run_analysis(checkpoint=checkpoint)                                                                                                        │
│   1218                                                                                                                                                │
│   1219                                                                                                                                                │
│   1220 if __name__ == "__main__":                                                                                                                     │
│                                                                                                                                                       │
│ /opt/venv/lib/python3.12/site-packages/cli/main.py:956 in run_analysis                                                                                │
│                                                                                                                                                       │
│    953 │   selected_analyst_keys = [a for a in ANALYST_ORDER if a in selected_set]                                                                    │
│    954 │                                                                                                                                              │
│    955 │   # Initialize the graph with callbacks bound to LLMs                                                                                        │
│ ❱  956 │   graph = TradingAgentsGraph(                                                                                                                │
│    957 │   │   selected_analyst_keys,                                                                                                                 │
│    958 │   │   config=config,                                                                                                                         │
│    959 │   │   debug=True,                                                                                                                            │
│                                                                                                                                                       │
│ /opt/venv/lib/python3.12/site-packages/tradingagents/graph/trading_graph.py:75 in __init__                                                            │
│                                                                                                                                                       │
│    72 │   │   set_config(self.config)                                                                                                                 │
│    73 │   │                                                                                                                                           │
│    74 │   │   # Create necessary directories                                                                                                          │
│ ❱  75 │   │   os.makedirs(self.config["data_cache_dir"], exist_ok=True)                                                                               │
[...]

PR #580, opened 2026-04-20
Add trading agents directory for appuser

kakumanivrn:
Create a directory for trading agents and set ownership to fix the `PermissionError: [Errno 13]` (#519)

PR #690, opened 2026-05-03
Fix Docker permission error for .tradingagents directories

cdavis-code:
## Summary

Fixes issue #627 - PermissionError when running TradingAgents in Docker.

## Problem

When running the application in Docker, users encountered a `PermissionError: [Errno 13] Permission denied: '/home/appuser/.tradingagents/cache'` error. This occurred because:

1. The Dockerfile switches to `appuser` with the `USER appuser` directive
2. The application tries to create directories at runtime using `os.makedirs()`
3. The `appuser` doesn't have permission to create directories in paths that don't exist

## Solution

Pre-create the required directories during the Docker build process and set proper ownership before switching to `appuser`:

- `/home/appuser/.tradingagents/cache`
- `/home/appuser/.tradingagents/results`  
- `/home/appuser/app/reports`

This ensures the directories exist with correct ownership when the application runs, preventing the permission error.

## Changes

- Modified Dockerfile to create necessary directories in the `RUN useradd` command
- Added `chown` to set proper ownership for `appuser`
- Directories are created before the `USER appuser` directive takes effect

## Testing

This fix resolves the error reported in issue #627 where users selecting "Enable Thinking" mode encountered the permission denied error during graph initialization.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,9 @@ ENV PYTHONDONTWRITEBYTECODE=1 \
 COPY --from=builder /opt/venv /opt/venv
 ENV PATH="/opt/venv/bin:$PATH"
 
-RUN useradd --create-home appuser
+RUN useradd --create-home appuser \
+ && mkdir -p /home/appuser/.tradingagents/cache /home/appuser/.tradingagents/results /home/appuser/app/reports \
+ && chown -R appuser:appuser /home/appuser/.tradingagents && chown -R appuser:appuser /home/appuser/app/reports
 USER appuser
 WORKDIR /home/appuser/app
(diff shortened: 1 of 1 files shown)


gliptak:
duplicate of #580

Yijia-Xiao:
Thanks @cdavis-code! Fixed in 704b762 via install -d (atomic create-with-ownership). We scoped it to /home/appuser/.tradingagents only — the runtime app creates cache/logs/memory subdirs at write time, and chowning the explicit .../results and .../reports paths isn't needed (results_dir default is actually .tradingagents/logs, not .tradingagents/results).

PR #698, opened 2026-05-03
fix(security): allow '=' character in ticker symbol for path generation

vinhmdev:
## Description
This PR addresses an issue where valid financial tickers containing the `=` character were being incorrectly rejected by the `safe_ticker_component` validator, which caused errors during filesystem operations (such as path generation for caching or reports).

Many Yahoo Finance tickers for Forex and Commodities rely on the `=` character (e.g., `XAUUSD=X` for Gold/USD, `EURUSD=X`, or `GC=F` for Gold Futures). This PR adds the `=` character to the `_TICKER_PATH_RE` regex to safely allow these common tickers to be processed while still maintaining protection against path traversal.

## Changes
- Updated `_TICKER_PATH_RE` in `tradingagents/dataflows/utils.py` to include the `=` character.
- Added test cases for `XAUUSD=X` and `GC=F` in `tests/test_safe_ticker_component.py` to ensure they pass validation safely.

### DIFF
--- a/tests/test_safe_ticker_component.py
+++ b/tests/test_safe_ticker_component.py
@@ -11,7 +11,7 @@ from tradingagents.dataflows.utils import safe_ticker_component
 @pytest.mark.unit
 class TestSafeTickerComponent(unittest.TestCase):
     def test_accepts_common_ticker_formats(self):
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @vinhmdev. The ticker path validator already allows '=' (futures like GC=F). Closing as resolved.

PR #713, opened 2026-05-04
Fix/ollama docker connection

### LINKED ISSUE #664: Connection failed error when trying to use local Ollama.
When running:
docker compose --profile ollama run --rm tradingagents-ollama

I get this error:
APIConnectionError: Connection error.
[NOTE] During task with name 'Market Analyst' and id '435ed680-925e-5fcf-e35d-45ff5293fb90'

I have tried on a fresh install of docker (on a different computer) and still got the error even though the image seemed to pull down fine and I'm able to go through the initial setup screens for TradingAgents.  As soon as it tries to "do something" after selecting options, it fails.

I have a server that runs ollama w/ a GPU that is separate from my desktop and I couldn't get it to work there either.  Same error.

AyushKar2005:
Fixes #664

## Problem
so basically when you run this inside docker, the ollama URL is hardcoded
to localhost:11434 which doesnt work. localhost inside a container points 
to the container itself not the host machine, so the connection just fails 
everytime. took me a bit to figure out why it wasnt connecting lol

## Changes
- `docker-compose.yml` - added ollama service with a healthcheck and a 
  separate tradingagents-ollama profile for people who want to run ollama 
  via docker
- `default_config.py` - now reads OLLAMA_BASE_URL from env variable, falls 
  back to localhost:11434 so local dev still works as before
- `openai_client.py` - same fix, was also hardcoded there
- `.env.example` - added the OLLAMA_BASE_URL variable so people know it exists

## How to test
spin it up with the ollama profile:

docker compose --profile ollama up

ollama should now be reachable at http://ollama:11434 inside the container 
network instead of trying to hit localhost and failing


Yijia-Xiao:
Thanks @AyushKar2005. The OLLAMA_BASE_URL plumbing you proposed for the Docker case is now first-party as of f10daa2 — set `OLLAMA_BASE_URL=http://<service-name>:11434/v1` (e.g. `http://ollama:11434/v1` when running both containers under the same docker-compose) and the value flows through both the CLI and programmatic paths. Closing this PR since the core env-var routing is shipped; the docker-compose healthcheck addition is a nice idea we'd want to revisit as its own focused change.

PR #728, opened 2026-05-05
fix: load CLI dotenv files from current working directory

### LINKED ISSUE #726: cli: load_dotenv() doesn't find user's .env when launched via the installed `tradingagents` console script
## Problem

When TradingAgents is installed via `pip install .` and launched with the `tradingagents` console script, environment variables in the user's `.env` file are not loaded — even when running from the repo root.

This causes confusing failures like:

> `OpenAIError: Missing credentials. Please pass an api_key... or set OPENAI_API_KEY...`

…even though the user's `.env` correctly contains the right provider key (e.g. `OPENROUTER_API_KEY`).

## Root cause

`cli/main.py` calls:

```python
load_dotenv()
load_dotenv(".env.enterprise", override=False)
```

With no `dotenv_path`, `python-dotenv`'s `load_dotenv()` falls back to `find_dotenv(usecwd=False)`, which walks **up from the calling source file's directory**, not from the user's CWD.

When the package is installed, `cli/main.py` lives at `…/site-packages/cli/main.py`. `find_dotenv` walks up through `site-packages → lib → python3.13 → envs → miniconda → …` — none of which contain a `.env` — and gives up. The user's project-level `.env` is never read.

This works fine when running from the repo source tree (`python -m cli.main`) because then `cli/main.py` is inside the repo and `find_dotenv` reaches the project `.env`. But the documented install path (`pip install .` then `tradingagents`) is silently broken for `.env`-based config.

## Fix

Use `find_dotenv(usecwd=True)` so the search starts from the user's working directory:

```python
from dotenv import load_dotenv, find_dotenv

load_dotenv(find_dotenv(".env.enterprise", usecwd=True), override=False)
```

Two-line change in `cli/main.py`. I've been running it locally without issues — happy to send a PR if useful.

## Reproduction

1. `git clone` the repo, `pip install .`
2. From the repo root: `cp .env.example .env` and add a real `OPENROUTER_API_KEY=...`
3. Run `tradingagents`, pick OpenRouter, any model
4. Crashes at graph init with the OpenAI "Missing credentials" error

## Environment

- macOS 15.x (Darwin arm64)
- Python 3.13.13 in a fresh conda env
- TradingAgents v0.2.4 (from `main`)
- `python-dotenv` (whatever version `pip install .` resolves)

Bortlesboat:
## Summary
- load `.env` and `.env.enterprise` by searching from the user's current working directory
- add a regression test that imports the installed-console-script entry module with dotenv calls stubbed

## Why
`tradingagents` is exposed as a console script, so `load_dotenv()` without an explicit path can search from the installed `cli/main.py` location instead of the directory where the user launched the command. That misses project-level `.env` files even when the user is in the repo or working directory that contains them.

Fixes #726.

## Verification
- RED: `python -m pytest tests/test_cli_env_loading.py -q` failed because dotenv was loaded without `find_dotenv(usecwd=True)`
- GREEN: `python -m pytest tests/test_cli_env_loading.py -q` -> `1 passed`
- `python -m py_compile cli/main.py tests/test_cli_env_loading.py`
- `python -m black --check tests/test_cli_env_loading.py`

## Note
A broader nearby run with `tests/test_google_api_key.py` is blocked in this local environment before reaching this patch by an existing dependency mismatch: `ImportError: cannot import name 'RunTree' from 'langsmith'` through `langchain_core`.

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -4,11 +4,11 @@ import typer
 from pathlib import Path
 from functools import wraps
 from rich.console import Console
-from dotenv import load_dotenv
+from dotenv import find_dotenv, load_dotenv
 
 # Load environment variables
-load_dotenv()
-load_dotenv(".env.enterprise", override=False)
+load_dotenv(find_dotenv(usecwd=True))
+load_dotenv(find_dotenv(".env.enterprise", usecwd=True), override=False)
 from rich.panel import Panel
 from rich.spinner import Spinner
 from rich.live import Live
--- a/tests/test_cli_env_loading.py
+++ b/tests/test_cli_env_loading.py
@@ -1,0 +1,38 @@
+import importlib
+import sys
+import types
+
+
+def test_cli_dotenv_searches_from_current_working_directory(monkeypatch):
+    calls = []
+
+    def fake_find_dotenv(filename=".env", usecwd=False):
+        calls.append(("find", filename, usecwd))
+        return f"/cwd/{filename}"
+
+    def fake_load_dotenv(dotenv_path=None, override=False):
+        calls.append(("load", dotenv_path, override))
+        return True
+
+    import dotenv
+
+    monkeypatch.setattr(dotenv, "find_dotenv", fake_find_dotenv)
+    monkeypatch.setattr(dotenv, "load_dotenv", fake_load_dotenv)
+    graph_package = types.ModuleType("tradingagents.graph")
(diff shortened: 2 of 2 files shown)


Yijia-Xiao:
Thanks @Bortlesboat! The fix you proposed (find_dotenv(usecwd=True)) landed in db7e0a6. The unit test you added is well-scoped — feel free to reopen as a test-only PR if you'd like to add it on top.

PR #730, opened 2026-05-05
Xiaomi Mimo Model Support

m-prabhu:
Added inbuilt support for Xiaomi Mimo Models


Yijia-Xiao:
Thanks @m-prabhu for the Xiaomi MiMo proposal. MiMo is typically served via vLLM or a similar OpenAI-compatible inference layer, so it's reachable today: pick `OpenAI` in the CLI and set `TRADINGAGENTS_LLM_BACKEND_URL` to your serving endpoint. Shipped d13e9b7. Closing as redundant — happy to revisit if Xiaomi exposes a hosted endpoint that requires special handling.

PR #747, opened 2026-05-06
fix: load .env using find_dotenv with usecwd=True (issue #726)

djconnexion77:
Fixes issue #726: environment variables were not loaded when the CLI is executed via the installed console script. Updated cli/main.py to use find_dotenv(usecwd=True) for both .env and .env.enterprise files.


Yijia-Xiao:
Thanks @djconnexion77 — your diagnosis was exactly right and the find_dotenv(usecwd=True) approach is what landed in db7e0a6. Both .env and .env.enterprise lookups now start from the user's CWD. The SMTP/email additions to .env.example you also included are a separate concern; happy to take those as their own PR if you'd like to open one.

PR #753, opened 2026-05-06
fix: load_dotenv searches CWD when launched via console script

### LINKED ISSUE #726: cli: load_dotenv() doesn't find user's .env when launched via the installed `tradingagents` console script

AyushKar2005:
## Problem

Fixes #726

so basically when you install tradingagents as a package and run it 
via the console script (just typing `tradingagents` in terminal), the 
`.env` file never gets picked up. all your API keys just dont load and 
it feels like the .env doesnt exist at all.

## Why this happens

turns out `load_dotenv()` by default looks for `.env` relative to 
where the script is, not where you are in the terminal. when you run 
`python cli/main.py` directly from the project folder it works fine 
because the script is right there. but once its installed as a package 
the script lives somewhere deep in site-packages so it never finds 
your .env

## Fix

changed `load_dotenv()` to `load_dotenv(find_dotenv(usecwd=True))` in 
both files. the `usecwd=True` part makes it search from whatever 
directory the user is actually standing in when they run the command, 
which is what you'd expect it to do in the first place honestly.

## Files changed

- `cli/main.py` — fixed both load_dotenv calls (the normal one and the .env.enterprise one)
- `main.py` — fixed the load_dotenv call there too

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -4,11 +4,11 @@ import typer
 from pathlib import Path
 from functools import wraps
 from rich.console import Console
-from dotenv import load_dotenv
+from dotenv import load_dotenv, find_dotenv
 
 # Load environment variables
-load_dotenv()
-load_dotenv(".env.enterprise", override=False)
+load_dotenv(find_dotenv(usecwd=True))
+load_dotenv(find_dotenv(usecwd=True, filename=".env.enterprise"), override=False)
 from rich.panel import Panel
 from rich.spinner import Spinner
 from rich.live import Live
--- a/main.py
+++ b/main.py
@@ -1,31 +1,22 @@
 from tradingagents.graph.trading_graph import TradingAgentsGraph
 from tradingagents.default_config import DEFAULT_CONFIG
+from dotenv import load_dotenv, find_dotenv
 
-from dotenv import load_dotenv
+load_dotenv(find_dotenv(usecwd=True))
 
-# Load environment variables from .env file
-load_dotenv()
-
-# Create a custom config
 config = DEFAULT_CONFIG.copy()
-config["deep_think_llm"] = "gpt-5.4-mini"  # Use a different model
-config["quick_think_llm"] = "gpt-5.4-mini"  # Use a different model
-config["max_debate_rounds"] = 1  # Increase debate rounds
+config["llm_provider"] = "openrouter"
+config["deep_think_llm"] = "meta-llama/llama-3.3-70b-instruct:free"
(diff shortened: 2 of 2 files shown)


Yijia-Xiao:
Thanks @AyushKar2005! The dotenv fix landed in db7e0a6 via find_dotenv(usecwd=True). The main.py rewrite (switching to OpenRouter, condensing comments) was kept out so the existing example stays neutral about provider choice.

Issue #641, opened 2026-05-01
GOOGLE_API_KEY is not being used


cranaweeraproject:
I get this error when trying to run:
ValidationError: 1 validation error for NormalizedChatGoogleGenerativeAI
  Value error, API key required for Gemini Developer API. Provide api_key parameter or set GOOGLE_API_KEY/GEMINI_API_KEY environment variable. [type=value_error, input_value={'model': 
'gemini-3.1-pro...gh', 'model_kwargs': {}}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.13/v/value_error

I select Google for llm provider in the process.
I have GOOGLE_API_KEY set in .env file and due to the error message I aded GEMINI_API_KEY
The GOOGLE_API_KEY is generated in google ai studio and is set as active.
The .env file is in the main directory and I'm launching it out of the main directory.

any idea how to fix it?

sejal12345hamal-hash:
That error isn’t about your key being invalid — it’s about your app **not seeing the key at runtime**. In other words, your `.env` file isn’t actually being loaded into environment variables.

Here are the most common fixes (one of these will solve it):

---

## 🔧 1. Make sure `.env` is actually loaded

Just having a `.env` file is **not enough**. You must explicitly load it in your code.

### Python (most likely your case)

Install dotenv if you haven’t:

```bash
pip install python-dotenv
```

Then at the **very top of your script**:

```python
from dotenv import load_dotenv
import os

load_dotenv()

print(os.getenv("GOOGLE_API_KEY"))  # debug check
```

If this prints `None`, your `.env` is not being read.

---

## 🔧 2. Check `.env` format (very common mistake)

Your `.env` file must look **exactly like this**:

```
GOOGLE_API_KEY=your_actual_key_here
GEMINI_API_KEY=your_actual_key_here
```

❌ Wrong:

```
GOOGLE_API_KEY = "abc123"
```

✔️ Correct:

```
GOOGLE_API_KEY=abc123
```

No spaces, no quotes.

---

## 🔧 3. Ensure you're running from the correct directory

Even if your `.env` is in the root, your script might not be running from there.

Quick fix:

```python
from dotenv import load_dotenv
import os

load_dotenv(dotenv_path=".env")
```

Or give full path:

```python
load_dotenv(dotenv_path="/full/path/to/.env")
```

---

## 🔧 4. Try setting it manually (to confirm issue)

Run this in your terminal:

### Linux / Mac:

```bash
[...]

PR #761, opened 2026-05-07
fix: prefer repo dotenv for provider API keys

marcusglee11:
## Summary
- Prefer this checkout's `.env` provider API key over a stale exported shell value when constructing OpenAI-compatible provider clients.
- Preserve the existing environment fallback when the repo `.env` does not define the selected provider key.
- Add regression coverage for the DeepSeek stale-shell-key case, including the `OpenAIClient.get_llm()` wiring path.

## Why
`python-dotenv` defaults to `override=False`, so a long-lived shell with an old `DEEPSEEK_API_KEY` can silently beat the key configured in the repo `.env`. The resulting DeepSeek failure reports the stale key suffix, which is confusing when `.env` is correct.

## Test plan
- [x] `UV_PROJECT_ENVIRONMENT=.venv-wsl uv run --with pytest python -m pytest tests/test_deepseek_reasoning.py -q`
- [x] `UV_PROJECT_ENVIRONMENT=.venv-wsl DEEPSEEK_API_KEY='sk-invalid-demo-3f4A' uv run python - <<'PY' ... OpenAIClient('deepseek-v4-flash', provider='deepseek').get_llm().invoke('Reply with OK only.') ... PY`

Related to issue #641.

### DIFF
--- a/tests/test_deepseek_reasoning.py
+++ b/tests/test_deepseek_reasoning.py
@@ -19,8 +19,56 @@ from langchain_core.prompt_values import ChatPromptValue
 from tradingagents.llm_clients.openai_client import (
     DeepSeekChatOpenAI,
     NormalizedChatOpenAI,
+    OpenAIClient,
     _input_to_messages,
+    _provider_api_key,
 )
+
+
+# ---------------------------------------------------------------------------
+# Provider API-key precedence
+# ---------------------------------------------------------------------------
+
+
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Shipped — the console script loads `.env` from the working dir (db7e0a6) and loading is centralized so every entry point picks it up. Closing as superseded.

PR #764, opened 2026-05-07
fix: pass max_recur_limit from config to Propagator

the-cyber-doc:
Propagator was always instantiated with its default of 100, ignoring the "max_recur_limit" value in DEFAULT_CONFIG.
This caused a GraphRecursionError.


Yijia-Xiao:
Thanks @the-cyber-doc! Plumbing fix landed in 9482cae — Propagator now picks up max_recur_limit from the config, defaulting to 100 to match the existing constructor default.

PR #774, opened 2026-05-08
Add OPENAI_BASE_URL support for OpenAI-compatible gateways

jomeswang:
## Summary
- add `OPENAI_BASE_URL` support for native OpenAI provider requests
- prefill the OpenAI CLI flow from `OPENAI_BASE_URL` while still allowing an interactive override
- document the new environment variable in `.env.example` and the README
- add tests for both the CLI selection flow and the OpenAI client fallback behavior

## Why
This makes it easier to use OpenAI-compatible gateways and proxies without patching local config for every run.

## Testing
- `uv run --with pytest pytest /Users/wwcome/work/demo/TradingAgents/tests/test_cli_provider_selection.py /Users/wwcome/work/demo/TradingAgents/tests/test_default_config_env.py /Users/wwcome/work/demo/TradingAgents/tests/test_ticker_symbol_handling.py`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,6 @@
 # LLM Providers (set the one you use)
 OPENAI_API_KEY=
+OPENAI_BASE_URL=
 GOOGLE_API_KEY=
 ANTHROPIC_API_KEY=
 XAI_API_KEY=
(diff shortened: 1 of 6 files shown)


Yijia-Xiao:
Thanks @jomeswang — the env-var route for the OpenAI backend URL is now first-party. Set `TRADINGAGENTS_LLM_BACKEND_URL=https://your-gateway/v1` (and your gateway's key as `OPENAI_API_KEY`) and the value is overlaid onto DEFAULT_CONFIG at startup. Shipped in d13e9b7 (#602); see `.env.example`. Closing as superseded.

PR #780, opened 2026-05-09
fix(docker): pre-create ~/.tradingagents in image so named volume inh…

bforejt:
…erits appuser ownership

When docker compose first creates the tradingagents_data named volume on /home/appuser/.tradingagents, Docker initializes the mountpoint as root:root because the path doesn't exist in the image. The container runs as appuser and fails on os.makedirs(...cache...) with PermissionError [Errno 13].

Create the directory in the Dockerfile and chown to appuser before the USER switch, so the volume picks up the correct ownership on first mount.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,9 @@ ENV PYTHONDONTWRITEBYTECODE=1 \
 COPY --from=builder /opt/venv /opt/venv
 ENV PATH="/opt/venv/bin:$PATH"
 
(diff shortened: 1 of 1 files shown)


nik2208:
same as #771 and #773

Yijia-Xiao:
Thanks @bforejt! Fixed in 704b762 via install -d (atomic create-with-ownership instead of mkdir + chown -R).

PR #785, opened 2026-05-09
Fix/screen flicker 784

### LINKED ISSUE #784: screen flicker when running analysis backend
how to avoid the flicker shaking screen?

AyushKar2005:
## fix for the screen flicker issue

Closes #784

hey so i was looking at this issue and i think i found whats causing the flicker. basically the Live display was refreshing 4 times per second which is alot and without screen=True it was just redrawing by scrolling the terminal instead of replacing it cleanly, thats what causes the shaking effect.

so i made two small changes:
- added screen=True to the Live() call so it uses a alternate terminal buffer (this is the main fix tbh)
- also lowered the refresh rate from 4 to 2 per second, doesnt really affect how it looks but reduces unnecessary redraws

```python
# before
with Live(layout, refresh_per_second=4) as live:

# after
with Live(layout, refresh_per_second=2, screen=True) as live:
```

tested it on windows and the flickering went away. only one line changed in cli/main.py so shouldnt break anything. lmk if you want me to change something

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -4,11 +4,11 @@ import typer
 from pathlib import Path
 from functools import wraps
 from rich.console import Console
-from dotenv import load_dotenv
+from dotenv import load_dotenv, find_dotenv
 
 # Load environment variables
-load_dotenv()
-load_dotenv(".env.enterprise", override=False)
+load_dotenv(find_dotenv(usecwd=True))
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @AyushKar2005. Implemented in 6398951.

PR #788, opened 2026-05-10
fix(dataflows): avoid mutable nested config leakage across runs

swamy18:
## Problem

`dataflows/config.py` used shallow copies for configuration state. Nested mutable structures such as `data_vendors` and `tool_vendors` could unintentionally share references across callers and global state.

This could lead to accidental cross-run mutation and non-deterministic behavior when configs are modified after initialization or update operations.

## Root Cause

The previous implementation relied on `.copy()`, which only creates a shallow copy of the top-level dictionary and does not isolate nested mutable objects.

Additionally, `set_config()` accepted caller-owned nested dictionaries directly into shared config state.

## Solution

Replaced shallow copy operations with `deepcopy()` in:

* `initialize_config()`
* `set_config()`
* `get_config()`

This ensures configuration state is fully isolated and prevents nested mutable aliasing between callers and internal global state.

## Testing

Added unit tests to verify:

* `get_config()` returns isolated nested structures
* `set_config()` does not retain caller-owned nested references

Validation performed:

* targeted tests passed
* full test suite passed (`110 passed`)

## Backward Compatibility

No public API changes.

This change only improves configuration isolation and deterministic behavior by preventing accidental shared-state mutation.

### DIFF
--- a/tests/test_dataflows_config.py
+++ b/tests/test_dataflows_config.py
@@ -1,0 +1,39 @@
+import copy
+import unittest
+
+import pytest
+
+import tradingagents.default_config as default_config
+from tradingagents.dataflows.config import get_config, set_config
+
+
+@pytest.mark.unit
+class DataflowsConfigIsolationTests(unittest.TestCase):
+    def setUp(self):
+        # Reset to defaults for deterministic test behavior.
+        set_config(copy.deepcopy(default_config.DEFAULT_CONFIG))
+
+    def test_get_config_returns_deep_copy(self):
+        cfg = get_config()
+        cfg["data_vendors"]["core_stock_apis"] = "alpha_vantage"
+        cfg["tool_vendors"]["get_stock_data"] = "alpha_vantage"
+
+        fresh = get_config()
+        self.assertEqual(fresh["data_vendors"]["core_stock_apis"], "yfinance")
+        self.assertNotIn("get_stock_data", fresh["tool_vendors"])
+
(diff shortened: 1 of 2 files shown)


swamy18:
If there are any changes you'd like in this PR - implementation details, tests, naming, or overall approach - I’d be happy to update them quickly. Open to feedback and improvements wherever needed.

swamy18:
Implemented the requested follow-up changes:

* `set_config()` now reuses `initialize_config()` (removed duplicated initialization logic).
* Replaced shallow top-level update behavior with one-level nested merge semantics so partial nested updates preserve existing defaults.
* Preserved isolation guarantees by deep-copying incoming config and returning deep copies from `get_config()`.

Added targeted tests for:

* nested default preservation on partial updates
* one-level nested merge behavior across successive updates
* existing deepcopy isolation guarantees

Validation:

* `tests/test_dataflows_config.py`: 4 passed
* full suite: 112 passed, 1 warning, 42 subtests passed

swamy18:
All requested review feedback has been addressed and the test suite is passing cleanly.

Happy to make any additional adjustments if needed.

Yijia-Xiao:
Thanks @swamy18! Your fix landed in 9482cae. We adopted both the deepcopy isolation and the one-level dict merge (so partial set_config calls don't clobber sibling defaults), plus your test cases for caller-side mutation isolation and partial-update preservation.

PR #790, opened 2026-05-10
feat: write run_config.md alongside each analysis (#752)

### LINKED ISSUE #752: Add report with run configuration details
It would be useful to include a separate configuration report for each run with all of the configuration details used in the run.

In my particular use-case, I'm testing across various LLM families (i.e. OpenAI vs. Google).  It would be useful to be able to tell which configuration was used to generate a particular set of reports for a run, including the following information:

- Version of tradingagents used
- Tradingagents params (e.g. research depth, reasoning effort etc.)
- LLM model name/s, versions and all parameters passed (e.g. prompt, temperature, other params etc.)
- Number of debate rounds

This would allow folks to determine the best configuration for their use case, and could also be used to prompt for re-run timing when newer models and TA software versions are released.

djconnexion77:
Closes #752.

Each saved analysis now includes a `run_config.md` at the report root, capturing:

- `tradingagents` package version (exposed via `importlib.metadata`)
- Ticker, analysis date, selected analysts, research depth, output language
- LLM provider, quick-thinking model, deep-thinking model, backend URL
- Max debate rounds, max risk discuss rounds, checkpoint flag
- Provider-specific reasoning params (OpenAI reasoning effort, Google thinking level, Anthropic effort) — only rendered when set

## Why

Issue #752 describes comparing runs across LLM families (OpenAI vs Google vs Ollama) and over time as models and tradingagents itself evolve. Without persisted config, you can't tell from disk which parameters produced which report.

## Implementation

- New `tradingagents.__version__` via `importlib.metadata.version("tradingagents")`
- New `cli/run_config_report.py` with one pure function: `build_run_config_markdown(selections, config, version) -> str`
- `cli/main.py:save_report_to_disk()` gains optional `selections`/`config` kwargs; when both are passed, writes `run_config.md`. Backwards compatible — existing callers that don't pass them still work.
- 3 unit tests in `tests/test_run_config_report.py` (`@pytest.mark.unit`)
- CHANGELOG updated under `[Unreleased] / Added`

## Sample output

```markdown
# Run Configuration

| Field | Value |
|-------|-------|
| tradingagents version | 0.2.4 |
| Ticker | NVDA |
| Analysis date | 2026-05-04 |
| Analysts | market, news, fundamentals |
| Research depth | 1 |
| LLM provider | ollama |
| Quick-thinking model | qwen3:latest |
| Deep-thinking model | gpt-oss:latest |
| Backend URL | http://localhost:11434/v1 |
| Max debate rounds | 1 |
| Max risk discuss rounds | 1 |
| Checkpoint enabled | False |
```

## Tests

```
$ python -m pytest -m unit
====================== 48 passed, 63 deselected in 1.24s =======================
```

(45 existing + 3 new for the markdown builder.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 Breaking changes within the 0.x line are called out explicitly.
 
+## [Unreleased]
+
+### Added
+- **Run configuration report.** Each saved analysis now includes `run_config.md`
+  alongside `complete_report.md`, capturing the tradingagents version, ticker,
+  analysis date, selected analysts, research depth, LLM provider and models,
+  debate/risk round counts, and any provider-specific reasoning parameters
+  (OpenAI reasoning effort, Google thinking level, Anthropic effort) that are
+  set. Useful for comparing runs across providers, models, and tradingagents
+  versions. Also exposes `tradingagents.__version__` via `importlib.metadata`. (#752)
+
 ## [0.2.4] — 2026-04-25
 
 ### Added
--- a/cli/main.py
+++ b/cli/main.py
@@ -638,8 +638,19 @@ def get_analysis_date():
             )
 
 
-def save_report_to_disk(final_state, ticker: str, save_path: Path):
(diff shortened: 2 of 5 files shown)


Yijia-Xiao:
Thanks @djconnexion77. Implemented in a1b3b5b as a header in complete_report.md and a run_settings entry in the state log, rather than a separate file.

PR #801, opened 2026-05-11
fix: some agents do not use the specified language when generating reports

liximin:


Yijia-Xiao:
Thanks @liximin — you correctly identified that several agents weren't honouring the configured output language. That structural fix has since landed on main (the researchers, risk debators, research manager, and trader all apply the language instruction now), so this PR's changes are already present. To make sure it can't regress again, cbc5f67d42635452b0d72015c4d1c1c375b50817 adds a guard test asserting every report-producing agent applies it. Appreciate you flagging it.

PR #820, opened 2026-05-12
fix: inject resolved instrument identity into prompts

### LINKED ISSUE #814: Bug: Multi-agent pipeline hallucinates wrong company identity — all analysts except fundamentals report on the wrong ticker
## Summary

When analyzing TOTDY (TOTO LTD. ADR — Japanese bathroom fixtures, ~$7.6B market cap), 5 of 6 agents in the pipeline analyzed **TotalEnergies SE** instead. Only the fundamentals analyst got it right. The error cascaded through the entire downstream chain (bull/bear debaters, trader, risk analysts, portfolio manager), producing a 1,500-line report about an oil supply crisis that has nothing to do with the actual company.

## Root Cause

`build_instrument_context()` only tells agents to use the exact ticker in tool calls — it does **not** inject the company name or identity:

```python
def build_instrument_context(ticker: str) -> str:
    return (
        f"The instrument to analyze is `{ticker}`. "
        "Use this exact ticker in every tool call, report, and recommendation, "
        "preserving any exchange suffix (e.g. `.TO`, `.L`, `.HK`, `.T`)."
    )
```

The market analyst gets correct price data from `get_stock_data("TOTDY")`, sees a parabolic chart, and the LLM **pattern-matches it to a narrative**: "oil stock in supply crisis" → hallucinates "TotalEnergies ADR" as the company name. Once the first analyst produces a report titled "#📊 TOTDY (TotalEnergies ADR)", every downstream agent follows that framing.

There is no **deterministic ticker-to-company-name resolution step** anywhere in the graph before the analysts start running.

## Evidence from agent logs

1. **Market analyst** — got correct price data but labeled it: `# 📊 TOTDY (TotalEnergies ADR) — Comprehensive Technical Analysis Report`

2. **News analyst** — actually found the correct TOTO LTD article ("TOTO is Up 20.2% After Strong FY2026 Earnings") but then **ignored it** and searched for "TotalEnergies" and "TTE" instead (both returned nothing)

3. **Sentiment analyst** — followed the market analyst's framing, reported on the Strait of Hormuz crisis, oil prices, windfall taxes

4. **Fundamentals analyst** — **only correct agent**: correctly reported `TOTDY (TOTO LTD.)` with accurate financials

5. **Downstream debaters/trader/risk/PM** — all debated TotalEnergies as if it were TOTDY, even arguing about whether "the ticker confusion" was a bug (the conservative analyst spotted the mismatch and they debated it at length without resolving it)

## Steps to reproduce

1. Run the default pipeline with `TOTDY` as the ticker
2. The market analyst will see a chart and likely misidentify the company
3. Cascade from there

(This is likely reproducible with any ticker where the price chart suggests a different industry than what the company actually is — especially OTC ADRs with non-obvious tickers.)

## Proposed fix

Add a **company name resolution step** that runs *before* any analyst begins:

1. Fetch the company name from yfinance (or the fundamentals data source) during graph initialization:
   ```python
   import yfinance as yf
   info = yf.Ticker(ticker).info
   company_name = info.get('longName') or info.get('shortName') or ticker
   sector = info.get('sector', 'unknown')
   ```

2. Inject the resolved identity into `build_instrument_context` so every agent's prompt includes:
   ```
   The instrument to analyze is `TOTDY`.
   **Company:** TOTO LTD. (Building Products & Equipment)
   ```

3. This is a deterministic, non-LLM lookup — yfinance already has this data for virtually all tickers. It should run in `create_initial_state()` or `propagate()` before graph execution.

## Version

Running latest main (v0.2.5 tag pulled today). No custom configuration — default pipeline with all four analyst types enabled.

## Related

- PR #312 previously fixed a different ticker-mapping bug (hardcoded dict) but the LLM hallucination path remains unaddressed.

EvanWangZ:
## Summary
- Resolve ticker identity once at graph startup using yfinance metadata
- Store the resulting instrument context in the initial agent state
- Reuse that context across analysts, researchers, trader, risk debaters, research manager, and portfolio manager prompts
- Add regression tests for resolved identity context and fail-open identity resolution

Resolves #814

## Testing
- `uv run --with pytest python -m pytest tests/test_ticker_symbol_handling.py tests/test_memory_log.py::TestPortfolioManagerInjection -q`
- `uv run --with pytest python -m pytest -q`

### DIFF
--- a/tests/test_memory_log.py
+++ b/tests/test_memory_log.py
@@ -674,6 +674,23 @@ class TestPortfolioManagerInjection:
         state = propagator.create_initial_state("NVDA", "2026-01-10")
         assert state["past_context"] == ""
 
+    def test_instrument_context_in_initial_state(self):
+        propagator = Propagator()
+        state = propagator.create_initial_state(
+            "TOTDY",
+            "2026-01-10",
+            instrument_context="Resolved instrument identity: Company: TOTO LTD.",
+        )
+        assert "instrument_context" in state
+        assert state["instrument_context"] == (
+            "Resolved instrument identity: Company: TOTO LTD."
+        )
+
+    def test_instrument_context_defaults_to_empty(self):
+        propagator = Propagator()
+        state = propagator.create_initial_state("NVDA", "2026-01-10")
+        assert state["instrument_context"] == ""
+
     # PM prompt
 
     def test_pm_prompt_includes_past_context(self):
--- a/tests/test_ticker_symbol_handling.py
+++ b/tests/test_ticker_symbol_handling.py
@@ -13,6 +13,9 @@ from tradingagents.agents.utils.agent_utils import (
 
 @pytest.mark.unit
 class TickerSymbolHandlingTests(unittest.TestCase):
+    def setUp(self):
+        resolve_instrument_identity.cache_clear()
+
     def test_normalize_ticker_symbol_preserves_exchange_suffix(self):
         self.assertEqual(normalize_ticker_symbol(" cnc.to "), "CNC.TO")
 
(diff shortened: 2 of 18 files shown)


EvanWangZ:
Addressed the review suggestion in `d574296`: added `@functools.lru_cache(maxsize=128)` to `resolve_instrument_identity` and added a regression test to verify repeated same-ticker lookups reuse the cached result.

Testing:
- `uv run --with pytest python -m pytest tests/test_ticker_symbol_handling.py tests/test_memory_log.py::TestPortfolioManagerInjection -q`
- `uv run --with pytest python -m pytest -q`

Yijia-Xiao:
Thanks @Zavianx — we implemented #814 in d7b40a2 built on your architecture (resolve identity via yfinance, store on state, consume across all agents). We extended it to also resolve on the CLI entry point — which builds the initial state directly rather than via propagate(), so it wasn't covered here — and harmonized build_instrument_context with the asset_type changes that landed on main since. Closing in favor of the shipped fix; appreciate the clean groundwork.

PR #836, opened 2026-05-14
fix(analysts): move current_date to top of system prompt

ms32035:
## Problem

Ollama models (and small LLMs generally) default to their training-cutoff date when generating `start_date`/`end_date` arguments for tool calls (`get_stock_data`, `get_news`, `get_global_news`, etc.).

The date hint was buried at the end of a 500+ word system message:
```
"For your reference, the current date is {current_date}."
```
By the time the model processes this, its training-weight prior for "now" has already taken hold.

## Fix

Move the date instruction to the very first line of every analyst system message and use directive language:
```
"Today's date is {current_date}. Use this date for all analysis and tool calls."
```

This ensures the model reads the correct date before any other instructions, overriding its internal sense of "now" when generating tool call arguments.

## Affected files

- `tradingagents/agents/analysts/market_analyst.py`
- `tradingagents/agents/analysts/fundamentals_analyst.py`
- `tradingagents/agents/analysts/news_analyst.py`
- `tradingagents/agents/analysts/sentiment_analyst.py`


Yijia-Xiao:
Thanks @ms32035 for catching this. Shipped in `2b2d685`: the current date now leads each analyst system prompt instead of trailing a long indicator block, so tool-call date ranges anchor correctly.

PR #839, opened 2026-05-15
Skip Anthropic effort for unsupported Claude models

### LINKED ISSUE #831: Anthropic effort parameter sent to Haiku 4.5 (which doesn't support it) → 400 error when using Haiku as quick-think model

haosenwang1018:
## Issue
Closes #831

## Root cause
`anthropic_effort` is currently forwarded to every Anthropic model. Mixed setups that use Haiku as the quick-think model and Opus/Sonnet as the deep-think model therefore send `effort` to Haiku, which rejects the parameter.

## Fix
Add a small Anthropic model capability check before forwarding `effort`. Opus 4.5/4.6/4.7, Sonnet 4.6, and Mythos Preview keep receiving the configured effort value; unsupported models such as Haiku 4.5 omit it while preserving other kwargs.

## Tests
- `uv run --with-editable . --with pytest python -m pytest tests/test_anthropic_effort.py tests/test_model_validation.py`

### DIFF
--- a/tests/test_anthropic_effort.py
+++ b/tests/test_anthropic_effort.py
@@ -1,0 +1,48 @@
+import pytest
+
+from tradingagents.llm_clients import anthropic_client as mod
+
+
+@pytest.mark.unit
+def test_haiku_does_not_receive_effort(monkeypatch):
+    captured = {}
+    monkeypatch.setattr(
+        mod,
+        "NormalizedChatAnthropic",
+        lambda **kwargs: captured.setdefault("kwargs", kwargs),
+    )
+
+    client = mod.AnthropicClient(
+        model="claude-haiku-4-5", effort="medium", api_key="placeholder"
+    )
+    client.get_llm()
+
+    assert "effort" not in captured["kwargs"]
+    assert captured["kwargs"]["api_key"] == "placeholder"
+
+
+@pytest.mark.unit
+@pytest.mark.parametrize(
+    "model",
+    [
+        "claude-opus-4-7",
+        "claude-opus-4-6",
+        "claude-opus-4-5",
+        "claude-sonnet-4-6",
+    ],
+)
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @haosenwang1018 — adopted your design in 61522e1. Added Sonnet 4.5 to the supported set and a forward-compat regex for future `claude-{opus,sonnet}-X-Y` releases. Test coverage expanded to cover the forward-compat path and unknown-model conservatism. Closing in favor of the shipped fix.

PR #847, opened 2026-05-17
fix(trading-agents-service): CORS for browser EventSource on /stream/{run_id}

jcoles1155:
## What

Dashboard's `<LiveStream>` component (EventSource → `/stream/{run_id}`) was failing CORS check from `https://lyceum-fund-app.vercel.app` because trading-agents-service has no CORS middleware. Error in browser console:

> Access to resource at `https://trading-agents-service-production-a6bc.up.railway.app/stream/...` from origin `https://lyceum-fund-app.vercel.app` has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.

## Why CORS didn't matter until now

- `POST /analyze` is hit by the Node-side worker (server-to-server). No browser, no CORS check, has been working since TT-283.
- `GET /stream/{run_id}` is hit by the browser via EventSource. CORS check enforced. Has been silently failing since TT-283 — we just never noticed because end-to-end test always loaded a completed run on a polled `getRun` query, never an in-flight one.

## Fix

Add Starlette's `CORSMiddleware` to `app/main.py`. Allow-list comes from `CORS_ALLOW_ORIGINS` env var (comma-separated) with a default of `https://lyceum-fund-app.vercel.app,http://localhost:3000`.

`allow_credentials=False` because the stream-token query-param pattern doesn't use cookies. `allow_methods=["GET", "POST", "OPTIONS"]` to cover the SSE GET and the future /analyze if anything browser-side ever needs it.

## Followups

- Vercel preview URLs (random `*.vercel.app` subdomains) won't match the allow-list. If we need previews to test SSE, switch to `allow_origin_regex=r"https://lyceum-fund-app.*\.vercel\.app"`. Defer until needed.
- The dashboard's "Failed to load run: Run not found" race (run row doesn't exist yet at the moment the dashboard navigates to it) is a separate UX gap — JC flagged it. Worth filing as polish.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,29 @@
+# ─────────────────────────────────────────────────────────────────────────
+# Two Trees service-wrapper vars (required by app/config.py)
+# ─────────────────────────────────────────────────────────────────────────
+
+# Neon Postgres — pooled URL. Shared with the Node lyceum-fund app.
+DATABASE_URL=
+# Neon Postgres — direct URL. Used by Alembic migrations.
+DIRECT_URL=
+# Upstash Redis URL. Shared with the Node side for pub-sub / SSE.
+REDIS_URL=
+# HMAC-SHA256 secret. Node-side worker signs requests; this service verifies.
+# Generate with: openssl rand -hex 32
+HMAC_SHARED_SECRET=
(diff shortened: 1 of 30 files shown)


Yijia-Xiao:

PR #848, opened 2026-05-17
feat(trading-agents-service): re-enable extractor via fire-and-forget asyncio task

jcoles1155:
## What

Re-enables the per-agent structured-data extractor (disabled in the previous PR's kill switch). Two changes that address the cross-loop bug at its root:

### 1. `on_chain_end` writes the report row, then SPAWNS the extractor

Previously, `on_chain_end` did:

write_row_with_extracted_metadata  ← awaited LLM call inline

Now it does:
write_row_with_null_metadata  ← returns immediately
asyncio.create_task(extract_and_update)  ← fire-and-forget

The LangChain callback dispatch finishes within a few ms instead of blocking for the LLM round-trip. The extractor task runs independently on the same loop and writes back to the row's metadata column when it completes.

A `_log_task_exception` done-callback surfaces unhandled task errors that would otherwise vanish silently.

### 2. `extract_metadata` no longer uses `asyncio.wait_for`

`asyncio.wait_for`'s timeout-cancellation path was the actual trigger for the "Future attached to a different loop" cascade. It cancelled the in-flight httpx coroutine in a way that left the asyncio context corrupted, which then poisoned subsequent operations on the singleton SQLAlchemy/asyncpg engine.

Now relies only on `ChatOpenAI(timeout=8)` — that's an httpx-level deadline, surfaces as a normal `TimeoutException`, no cancellation games.

## Why this should hold

The previous failure mode required two conditions to align:

1. `asyncio.wait_for` cancellation poisoning the loop context — gone (no more `wait_for`)
2. The poisoned context being used for DB ops via singletons — partially mitigated (the SQLAlchemy engine singleton remains, but it's never touched by the corrupted path now)

Even if some other latent corruption surfaces later, the fire-and-forget pattern prevents extractor failures from cascading into chain errors. The chain callback completes regardless of what the extractor does.

## Files

- `app/services/callbacks.py`:
  - New private methods `_persist_report` (writes row with null metadata, returns id) and `_extract_and_update` (fire-and-forget body)
  - `on_chain_end` orchestrates: persist row, spawn task, publish event
  - New module-level `_log_task_exception` callback for surfaced task failures
- `app/services/extractors/extractor.py`:
  - Removed `asyncio.wait_for` wrapper
  - Removed kill-switch early return
  - Simpler error handling (single catch-all)

## Smoke test after deploy

1. Submit a fresh AAPL or TSLA analysis
2. Watch logs: should see normal chain progression, no `attached to a different loop`, no runaway loops
3. Check the DB:
```sql
   SELECT agent_name, jsonb_pretty(metadata)
   FROM agent_reports
   WHERE run_id = '<runId>'
   ORDER BY created_at;
```
   Metadata should populate for the 12 agents in `SCHEMA_FOR_AGENT` (lower-cased: market_analyst, fundamentals_analyst, etc.) and stay null for utility nodes (tools_market, msg_clear_*, etc.).
4. Dashboard: expand any agent `<details>` — metadata card should render with structured data; prose below.

## Risk + escape hatch

If the cross-loop bug somehow returns: the kill switch pattern is well-documented; flipping the extractor back to `return None` is a one-line revert. But the underlying trigger (`asyncio.wait_for`) is gone, so this should hold.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,29 @@
+# ─────────────────────────────────────────────────────────────────────────
+# Two Trees service-wrapper vars (required by app/config.py)
+# ─────────────────────────────────────────────────────────────────────────
+
+# Neon Postgres — pooled URL. Shared with the Node lyceum-fund app.
+DATABASE_URL=
+# Neon Postgres — direct URL. Used by Alembic migrations.
+DIRECT_URL=
+# Upstash Redis URL. Shared with the Node side for pub-sub / SSE.
+REDIS_URL=
+# HMAC-SHA256 secret. Node-side worker signs requests; this service verifies.
+# Generate with: openssl rand -hex 32
+HMAC_SHARED_SECRET=
+# Slug of the spawned app this service backs. Tagged on every Sentry event.
+APP_SLUG=lyceum-fund
+# Sentry DSN. Errors flow to the shared two-trees-shared-python project.
+SENTRY_DSN=
+# 'production' on Railway; 'development' locally.
+NODE_ENV=development
+# Bind port. Railway injects automatically.
+PORT=8000
+
+# ─────────────────────────────────────────────────────────────────────────
+# Upstream TradingAgents env vars (consumed by tradingagents/ library)
+# ─────────────────────────────────────────────────────────────────────────
+
 # LLM Providers (set the one you use)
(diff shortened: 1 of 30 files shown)


Yijia-Xiao:

PR #849, opened 2026-05-17
fix(trading-agents-service): per-run DB engine — root-cause cross-loop bug

jcoles1155:
## What

The cross-loop bug that's been blocking us for the last several PRs has finally got an actual root-cause fix.

**The problem:** the global SQLAlchemy/asyncpg engine (`app/db.py` `get_engine()`) is a process-wide singleton. It's lazy-initialized on the first request that calls `get_sessionmaker()` — usually `/health` from Railway's probe. That binds the engine's connection pool to whatever event loop was current at that moment.

`run_analysis` runs as a FastAPI BackgroundTask, which executes in a **different event loop context**. Every DB write inside the callback handler (`on_chain_end` → `_persist_report` → `session.commit()`) is awaiting a Future that's bound to the wrong loop. Result: `got Future <Future pending cb=[BaseProtocol._on_waiter_completed()]> attached to a different loop`, which cascades into chain errors → retries → runaway loops.

Reverting singletons (Redis, ChatOpenAI) helped narrow it. The fire-and-forget extractor pattern still hit it. The kill switch hid the symptom but didn't fix the cause.

## Fix

`run_analysis` now creates its OWN engine + sessionmaker at the top, inside the BackgroundTask coroutine — so it's bound to **this run's loop**. Every downstream DB op (status updates, agent_reports persist via the callback, decision write, fire-and-forget extractor updates) routes through it.

```python
async def run_analysis(...):
    url, connect_args = asyncpg_url(settings.DATABASE_URL)
    engine = create_async_engine(url, connect_args=..., pool_size=5, max_overflow=10)
    sessionmaker = async_sessionmaker(engine, class_=AsyncSession, expire_on_commit=False)

    try:
        # ... existing pipeline ...
    finally:
        await publish_done(run_id)
        await engine.dispose()  # clean up connection pool
```

The `/health` and `/ready` routes continue using the global singleton (which is fine — they're on the request-handler loop). Only the BackgroundTask-spawned analysis gets a dedicated engine.

## Cost

1 TCP+TLS handshake to Neon per run (pool warmup). 5 connections shared across ~12 agents. Disposed cleanly at run-end. Negligible vs the cost of the analysis itself.

## Why the previous attempts didn't fix this

- **Reverting Redis/ChatOpenAI singletons** — addressed pub-sub + LLM-call cross-loop, but the DB engine is the bigger one
- **Fire-and-forget extractor** — the LangChain callback returned faster, but `_persist_report` still used the singleton engine → still cross-loop
- **Kill switch** — removed the extractor entirely; persist STILL failed sometimes, just with fewer total calls so it was less visible. The TSLA run that "seemed to work" probably did have some persist failures we didn't see scroll past

## Trade-offs

- Fire-and-forget extractor tasks spawned late in a run may not complete before `engine.dispose()` — those metadata writes are lost (logged via `_log_task_exception`). Acceptable: metadata is best-effort enrichment. The agent's prose content and the decision row are persisted before any extractor task is spawned.

## Smoke test after deploy

1. Submit a fresh AAPL analysis
2. Logs should be CLEAN — no "different loop", no asyncpg `InternalClientError`, no runaway market_analyst loop
3. Run completes, decision card renders
4. Agent reports populate with content + metadata
5. Dashboard metadata cards render per-agent (RSI gauge, sentiment scale, etc.)

If something still surfaces cross-loop after this — that would be deeper LangGraph weirdness and we'd need to investigate further. But this matches the most common asyncpg + FastAPI BackgroundTask pattern in the docs.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,29 @@
+# ─────────────────────────────────────────────────────────────────────────
+# Two Trees service-wrapper vars (required by app/config.py)
+# ─────────────────────────────────────────────────────────────────────────
+
+# Neon Postgres — pooled URL. Shared with the Node lyceum-fund app.
+DATABASE_URL=
+# Neon Postgres — direct URL. Used by Alembic migrations.
+DIRECT_URL=
+# Upstash Redis URL. Shared with the Node side for pub-sub / SSE.
+REDIS_URL=
+# HMAC-SHA256 secret. Node-side worker signs requests; this service verifies.
+# Generate with: openssl rand -hex 32
+HMAC_SHARED_SECRET=
+# Slug of the spawned app this service backs. Tagged on every Sentry event.
+APP_SLUG=lyceum-fund
+# Sentry DSN. Errors flow to the shared two-trees-shared-python project.
+SENTRY_DSN=
+# 'production' on Railway; 'development' locally.
+NODE_ENV=development
+# Bind port. Railway injects automatically.
+PORT=8000
+
+# ─────────────────────────────────────────────────────────────────────────
+# Upstream TradingAgents env vars (consumed by tradingagents/ library)
+# ─────────────────────────────────────────────────────────────────────────
+
 # LLM Providers (set the one you use)
 OPENAI_API_KEY=
 GOOGLE_API_KEY=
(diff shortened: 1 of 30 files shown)


Yijia-Xiao:

PR #850, opened 2026-05-17
fix(trading-agents-service): JSONB(none_as_null=True) — extractor never ran for any row

jcoles1155:
## What

The post-pipeline extractor from the previous PR was running successfully but **silently no-op'ing on every row**. SQL query confirmed:
`has_metadata = t` means `metadata IS NOT NULL` — the column has a value. The displayed `null` is the **JSON null literal** (`jsonb 'null'`), not SQL NULL.

## Root cause

SQLAlchemy's default JSONB column behavior: when you assign Python `None`, it writes the JSON `null` literal (because in JSON, `null` is a valid value distinct from "missing"). To get SQL NULL, you have to opt in with `none_as_null=True`.

So `on_chain_end`'s `report_metadata=None` was writing JSON null, not SQL NULL. Then `_extract_run_metadata`'s `WHERE report_metadata IS NULL` matched zero rows — extractor saw nothing to enrich.

## Fix

One word added to the column definition:

```python
report_metadata = Column("metadata", JSONB(none_as_null=True), nullable=True)
```

Now `None` → SQL NULL on writes. `IS NULL` matches correctly. Extractor sees the rows. Cards render.

## Smoke test after deploy

1. Submit a fresh AAPL or TSLA or NVDA analysis
2. Wait for status → `complete`
3. Run the SQL:
```sql
   SELECT agent_name,
          (metadata IS NOT NULL) AS has_metadata,
          jsonb_pretty(metadata) AS metadata
   FROM agent_reports
   WHERE run_id = '<new-runId>'
   ORDER BY created_at;
```
4. Expected: `has_metadata = true` for `market_analyst`, `fundamentals_analyst`, `news_analyst`, etc. with REAL JSON dicts; `has_metadata = false` (SQL NULL) for utility nodes like `tools_market`, `msg_clear_*` (no schema defined).
5. Dashboard: metadata cards render per-agent on the detail page.

## Note about existing runs (NVDA + earlier)

Rows from previous runs have JSON null literally stored. They'll continue to render no card (because `MetadataCard` correctly returns null for null metadata). New runs from here onward will have proper SQL NULL initially → extractor populates → cards render. Acceptable; the broken runs are test data.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,29 @@
+# ─────────────────────────────────────────────────────────────────────────
+# Two Trees service-wrapper vars (required by app/config.py)
+# ─────────────────────────────────────────────────────────────────────────
+
+# Neon Postgres — pooled URL. Shared with the Node lyceum-fund app.
+DATABASE_URL=
+# Neon Postgres — direct URL. Used by Alembic migrations.
+DIRECT_URL=
+# Upstash Redis URL. Shared with the Node side for pub-sub / SSE.
+REDIS_URL=
+# HMAC-SHA256 secret. Node-side worker signs requests; this service verifies.
+# Generate with: openssl rand -hex 32
+HMAC_SHARED_SECRET=
+# Slug of the spawned app this service backs. Tagged on every Sentry event.
+APP_SLUG=lyceum-fund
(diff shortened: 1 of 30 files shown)


Yijia-Xiao:

Issue #826, opened 2026-05-13
Bug: TypeError: Completions.create() got an unexpected keyword argument 'reasoning_split' with new MiniMax compatible API


malaccan:
## Description

When attempting to use the newly released MiniMax compatible API (global Minimax Coding Plan) via the OpenAI client wrapper, the API call fails with a `TypeError`. The underlying openai Python SDK strictly validates arguments and is rejecting the `reasoning_split` parameter, suggesting `reasoning_effort` instead.

## To Reproduce
1. Set up the openai client to point to the new MiniMax compatible endpoint.
2. Enter a Minimax Coding Plan API key.
3. Trigger an agent/chain that invokes the LLM (in this case, `market_analyst_node`).
4. The chain attempts to pass `reasoning_split` as a keyword argument to the completion creation method.
5. See error: `TypeError: Completions.create() got an unexpected keyword argument 'reasoning_split'. Did you mean 'reasoning_effort'`?

## Quick Fix
Tested a (dirty) fix by simply changing `reasoning_split` to `reasoning_effort` inside `openai_client.py` and the code works without issues for minimax coding plan.

## Error Traceback

```bash
╭──────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────╮
│ /Users/auser/python/random/trading_agents/TradingAgents/cli/main.py:1254 in analyze           │
│                                                                                                            │
│   1251 │   │   from tradingagents.graph.checkpointer import clear_all_checkpoints                          │
│   1252 │   │   n = clear_all_checkpoints(DEFAULT_CONFIG["data_cache_dir"])                                 │
│   1253 │   │   console.print(f"[yellow]Cleared {n} checkpoint(s).[/yellow]")                               │
│ ❱ 1254 │   run_analysis(checkpoint=checkpoint)                                                             │
│   1255                                                                                                     │
│   1256                                                                                                     │
│   1257 if __name__ == "__main__":                                                                          │
│                                                                                                            │
│ /Users/auser/python/random/trading_agents/TradingAgents/cli/main.py:1090 in run_analysis      │
│                                                                                                            │
│   1087 │   │                                                                                               │
│   1088 │   │   # Stream the analysis                                                                       │
│   1089 │   │   trace = []                                                                                  │
│ ❱ 1090 │   │   for chunk in graph.graph.stream(init_agent_state, **args):                                  │
│   1091 │   │   │   # Process all messages in chunk, deduplicating by message ID                            │
[...]

PR #860, opened 2026-05-19
fix(llm): pass MiniMax reasoning_split via extra_body

CaoYuhaoCarl:
## Summary
- MiniMax's `reasoning_split` flag was being placed at the top level of the request payload, where the openai SDK's strict kwarg validation rejects it.
- Moves it into `extra_body` so the SDK forwards it untouched, following the gating already added in #826.

## Test plan
- [ ] Call a MiniMax reasoning model (e.g. MiniMax-M2) and confirm `reasoning_split` reaches the API without an SDK validation error.
- [ ] Call a non-reasoning MiniMax endpoint (Coding Plan, MiniMax-Text-01) and confirm the flag is still gated off by `requires_reasoning_split`.


Yijia-Xiao:
Thanks @CaoYuhaoCarl — you're right, and it's shipped in 8694bd0. Our earlier fix gated reasoning_split by model but still passed it top-level, so M2.x models still hit the openai SDK's validation. Moved it into `extra_body` as you proposed. Closing in favor of the shipped fix.

PR #865, opened 2026-05-19
feat(docker): allow binding tradingagents_data volume to a host path

sjq597:
The compose stack persists reports, cache, and the decision log under /home/appuser/.tradingagents via a named volume. That volume is opaque to the host, so users running stock analyses through Docker have no direct way to read the generated reports under
logs/<TICKER>/<DATE>/reports/.

Make the mount source overridable through the TRADINGAGENTS_DATA_DIR environment variable (consumed by docker compose, not by default_config.py). Unset, behavior is unchanged: the named volume 'tradingagents_data' is still used. Set it to a host path (e.g. ./tradingagents_data) to bind-mount instead, exposing all reports on the host filesystem.

- docker-compose.yml: parameterize both service mounts with ${TRADINGAGENTS_DATA_DIR:-tradingagents_data}
- .env.example: document the new variable in its own section so it is not confused with the TRADINGAGENTS_* DEFAULT_CONFIG overlay
- .gitignore: ignore tradingagents_data/ for users who opt into the bind mount under the project root
- README.md: add a short note in the Docker section explaining how to switch to a host-visible directory

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,3 +30,11 @@ OPENROUTER_API_KEY=
 #TRADINGAGENTS_MAX_DEBATE_ROUNDS=1
 #TRADINGAGENTS_MAX_RISK_ROUNDS=1
 #TRADINGAGENTS_CHECKPOINT_ENABLED=false
+
+# Optional (Docker only): where to persist /home/appuser/.tradingagents
+# (reports, cache, decision log). Unset = use the named volume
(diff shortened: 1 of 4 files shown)


Yijia-Xiao:
Thanks @sjq597. Implemented in 45391d0: set TRADINGAGENTS_DATA_DIR to keep results, reports and the memory log in a host folder.

PR #868, opened 2026-05-20
fix: move reasoning_split into extra_body to fix openai SDK compatibi…

mihailnica10:
…lity

Minimax M2.x models require reasoning_split=True in the request body to redirect thinking blocks into reasoning_details. This worked with older openai SDK versions that didn't validate top-level params, but openai SDK >= 1.56 validates against Completions.create signature and rejects reasoning_split as unknown.

The fix puts reasoning_split in extra_body instead of at top-level, which is the correct location for provider-specific parameters and avoids the SDK validation error.


Yijia-Xiao:
Thanks @mihailnica10 — confirmed and shipped in 8694bd0: reasoning_split now goes via `extra_body` instead of top-level, so the openai SDK (>=1.56) no longer rejects it for M2.x models. Closing in favor of the shipped fix.

PR #876, opened 2026-05-21
docs: clarify Tushare and AkShare support

AyushKar2005:


Yijia-Xiao:
We analyze A-shares through the Yahoo Finance vendor (.SS/.SZ) rather than Tushare/AkShare in core, so this clarification doesn't apply. Closing.

PR #881, opened 2026-05-22
fix: verify market data before analyst reports

### LINKED ISSUE #830: Model hallucinating? How reliable is it?
Hi

I am analyzing COF ticker and Portfolio Manager report says `the Bollinger lower band at $181.30 provides a historically validated support level (12–15% bounces in March 2026 and October 2025)`, but  looking at price and Bollinger band for October 2025 it is around $200. This is obviously concerning if I can't trust this tool on basic things such as price data. Is this issue with the model or agent? Maybe there needs to be another agent to verify the market data. 

Ticker: COF
Analysis date: 2026-05-13  
Language: English
Analyst Team: market, social, news, fundamentals
Research depth: Medium
[Quick-Thinking LLM Engine]: DeepSeek V4 Flash - Latest V4 fast model
[Deep-Thinking LLM Engine]: DeepSeek V4 Pro - Latest V4 flagship model

Image

Image

AyushKar2005:
Fixes #830

This PR adds a deterministic market data verification step for the Market Analyst.

The issue was that the generated report could mention exact technical levels, like Bollinger Band values or historical support/bounce claims, but sometimes those numbers were not matching the actual market data. So instead of only relying on the LLM's written reasoning, I added a seperate verification snapshot tool which gives the agent a more grounded source of truth before it writes the final market report.

### What changed

- Added a new `market_data_validator.py` dataflow helper.
- Added `get_verified_market_snapshot` as a LangChain tool.
- Wired the new tool into the Market Analyst.
- Updated the Market Analyst prompt so it must call this verification tool before writing exact OHLCV / price-level / indicator-value claims.
- Added tests for the new verified market snapshot logic.

### What the new tool returns

The new snapshot includes:

- latest OHLCV row on or before the analysis date
- selected technical indicators like:
  - Bollinger Bands
  - RSI
  - MACD
  - moving averages
  - ATR
- recent verified closing prices for sanity checking

It also makes sure data after the requested analysis date is excluded, so the report does not accidently use future rows.

### Why this helps

This should reduce cases where the agent says things like a Bollinger lower band or support level is at some exact price, when the actual indicator data does not support that claim.

The prompt now tells the Market Analyst to treat this verified snapshot as the source of truth for exact market data claims. If some other tool output conflicts with it, the agent should explicitly mention the mismatch instead of making up a reconciled number.

This does not completely solve all hallucinations, but it adds a deterministic check around the part that should be factual: price data, indicator values, dates, and recent closes.

### Validation

I ran:

```bash
python -m compileall tradingagents
python -m pytest tests/test_market_data_validator.py -q

Result :
2 passed
```

### DIFF
--- a/tests/test_market_data_validator.py
+++ b/tests/test_market_data_validator.py
@@ -1,0 +1,68 @@
+from __future__ import annotations
+
+import pandas as pd
+
+import tradingagents.dataflows.market_data_validator as validator
+
+
+def _sample_ohlcv() -> pd.DataFrame:
+    dates = pd.bdate_range("2026-04-01", "2026-05-20")
+    closes = [100 + i for i in range(len(dates))]
+
+    return pd.DataFrame(
+        {
+            "Date": dates,
+            "Open": [value - 0.5 for value in closes],
+            "High": [value + 1.0 for value in closes],
+            "Low": [value - 1.0 for value in closes],
+            "Close": closes,
+            "Volume": [1_000_000 + i for i in range(len(dates))],
+        }
+    )
+
+
+def test_verified_market_snapshot_filters_future_rows(monkeypatch):
+    data = _sample_ohlcv()
+
+    future_row = pd.DataFrame(
+        {
+            "Date": [pd.Timestamp("2026-06-01")],
+            "Open": [999.0],
+            "High": [999.0],
+            "Low": [999.0],
+            "Close": [999.0],
+            "Volume": [999],
+        }
+    )
+
+    data = pd.concat([data, future_row], ignore_index=True)
+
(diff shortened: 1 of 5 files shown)


Yijia-Xiao:
Thanks @AyushKar2005 — shipped in 47cbb32 on your design (the verification-snapshot dataflow + `get_verified_market_snapshot` tool wired into the market analyst with a source-of-truth instruction). Rebased onto current main and simplified the snapshot builder to reuse `load_ohlcv`'s normalization. Verified live on the COF case from #830 — accurate numbers, no fabricated bounces. Closing in favor of the shipped change.

Issue #787, opened 2026-05-10
Gitignore .env.enterprise and reports folder


kunalk16:
Currently the .env.enterprise and the /reports folder is not ignored, we need to ignore them, especially the .env.enterprise file

kunalk16:
Fixing this with a no-brainer PR: https://github.com/TauricResearch/TradingAgents/pull/786

Adamchaua:
Hi! I can help with this. I'll update the .gitignore to include .env.enterprise and the /reports folder to ensure sensitive data and local reports aren't accidentally committed. Thanks!

kunalk16:
This is one small chore, there are 4 PRs already and maintainers seem unreachable.

kunalk16:
Closing as completed, maintainer themselves checked in to main branch.

PR #884, opened 2026-05-22
fix: ignore .env.enterprise and reports folder (#787)

Adamchaua:
Added .env.enterprise and reports/ to .gitignore as requested in #787. This prevents sensitive environment files and local analysis reports from being accidentally committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -149,6 +149,7 @@ activemq-data/
 
 # Environments
 .env
+.env.enterprise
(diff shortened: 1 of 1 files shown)


Yijia-Xiao:
Shipped in d6762d6 — `.env.enterprise` and `reports/` are now gitignored. Closing this and the other duplicate PRs for the same change. Thanks!

PR #894, opened 2026-05-25
fix: make analyst placeholder context-aware

### LINKED ISSUE #888: Bug: Analyst workflow can leak `Continue` as the next analyst prompt
### Summary

In the analyst workflow, after each analyst finishes, the graph clears the message history and inserts a placeholder `HumanMessage(content="Continue")`.

This can cause some LLM providers or OpenAI-compatible proxy providers to interpret `"Continue"` as the actual user task, instead of continuing the TradingAgents workflow for the selected ticker. As a result, later analysts may produce irrelevant output about the word "continue" rather than analyzing the instrument.

### Observed behavior

When running an analysis for ticker `EC` (`Ecopetrol`) using an OpenAI-compatible provider, some analyst reports contained responses like:

> Parece que tu mensaje actual es "Continue"...

or returned definitions/explanations of the word "continue" instead of performing the assigned financial analysis.

This polluted downstream reports and could affect the final trading decision.

### Expected behavior

After clearing messages, the next analyst should still be anchored to the actual workflow context:

- Ticker / instrument
- Asset type
- Analysis date
- Assigned analyst task

The placeholder should not be a bare natural-language command like `Continue`.

### Root cause

`create_msg_delete()` currently clears the message history and adds:

```python
placeholder = HumanMessage(content="Continue")
```

### Environment

- Project: TradingAgents
- Workflow: CLI
- Ticker tested: `EC`
- Company: Ecopetrol
- Provider: OpenAI-compatible provider / Pollinations-style proxy
- Output language tested: Spanish only

### Additional note

I have only reproduced this while running the analysis in Spanish. I have not yet verified whether the same issue appears with English output or other output languages.

That said, the root cause appears language-independent: after message clearing, the workflow inserts a bare `HumanMessage(content="Continue")`, which some providers/models may interpret literally as the full user request.

### Suggested fix

Replace the ambiguous `Continue` placeholder with a task-aware message that keeps the next analyst anchored to the actual workflow context.

Current behavior:

```python
placeholder = HumanMessage(content="Continue")
```

Suggested replacement:

```python
ticker = state.get("company_of_interest", "the requested instrument")
trade_date = state.get("trade_date", "the requested date")
asset_type = state.get("asset_type", "stock")

placeholder = HumanMessage(
    content=(
        "Proceed with your assigned analysis for this workflow. "
        f"The instrument to analyze is `{ticker}`. "
        "Use this exact ticker in every tool call, report, and recommendation. "
        f"The asset type is {asset_type}. "
        f"The analysis date is {trade_date}."
    )
)
```

This keeps the compatibility benefit of leaving a human message after clearing history, while avoiding a bare natural-language command that some models may interpret as the actual user request.

AyushKar2005:
Fixes #888

This PR fixes the issue where the analyst workflow could pass a bare `Continue` message to the next analyst after clearing the previous messages.

Earlier, `create_msg_delete()` was deleting the old messages and then adding this placeholder:

```python
HumanMessage(content="Continue")
```

This can be a problem for some OpenAI-compatible providers or proxy providers because they may treat `Continue` as the actual user prompt. Because of that, some later analyst reports can end up explaining the word "continue" or replying like the user only said continue, instead of doing the financial analysis for the selected ticker.

So I changed the placeholder to be more context-aware. It now keeps the next analyst anchored to the actual workflow context.

The new placeholder includes:

- the ticker / instrument
- the asset type
- the analysis date
- instruction to use the exact ticker in tool calls, reports and recommendations
- a note that this placeholder should not be treated as a standalone user request

This keeps the compatibility reason for having a HumanMessage after clearing history, but avoids sending a vague natural language command like `Continue`.

### What changed

- Updated `create_msg_delete()` in `agent_utils.py`
- Added a helper to build a safer workflow placeholder message
- Added tests for the placeholder behavior
- Verified that the placeholder is no longer just `Continue`
- Verified that ticker, asset type and trade date are included in the new placeholder

### Why this helps

This should prevent later analysts from losing the actual task context after message cleanup. It also makes the workflow safer for OpenAI-compatible providers that are more likely to interpret the last human message very literally.

It does not change the overall graph flow, only makes the placeholder message less ambiguous and more tied to the current TradingAgents workflow.

### Validation

I ran:

```bash
python -m compileall tradingagents/agents/utils/agent_utils.py
python -m pytest tests/test_agent_utils.py -q
```

Result:

```bash
2 passed
```

### DIFF
--- a/tests/test_agent_utils.py
+++ b/tests/test_agent_utils.py
@@ -1,0 +1,53 @@
+from langchain_core.messages import AIMessage, HumanMessage, RemoveMessage
+
+from tradingagents.agents.utils.agent_utils import create_msg_delete
+
+
+def test_create_msg_delete_uses_context_aware_placeholder():
+    delete_messages = create_msg_delete()
+
+    result = delete_messages(
+        {
+            "messages": [
+                HumanMessage(content="old user message", id="human-1"),
+                AIMessage(content="old ai message", id="ai-1"),
+            ],
+            "company_of_interest": "EC",
+            "trade_date": "2026-05-13",
+            "asset_type": "stock",
+        }
+    )
+
+    messages = result["messages"]
+    placeholder = messages[-1]
+
+    assert len(messages) == 3
+    assert all(isinstance(message, RemoveMessage) for message in messages[:-1])
+    assert isinstance(placeholder, HumanMessage)
+
+    assert placeholder.content != "Continue"
+    assert "`EC`" in placeholder.content
+    assert "stock" in placeholder.content
+    assert "2026-05-13" in placeholder.content
+    assert "assigned analysis" in placeholder.content
+    assert "standalone user request" in placeholder.content
+
+
+def test_create_msg_delete_placeholder_has_safe_defaults():
+    delete_messages = create_msg_delete()
+
+    result = delete_messages(
+        {
+            "messages": [
+                HumanMessage(content="old user message", id="human-1"),
+            ],
+        }
+    )
+
+    placeholder = result["messages"][-1]
+
+    assert isinstance(placeholder, HumanMessage)
+    assert placeholder.content != "Continue"
+    assert "the requested instrument" in placeholder.content
+    assert "the requested date" in placeholder.content
+    assert "stock" in placeholder.content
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @AyushKar2005 — the #888 fix shipped in d7b40a2. We took @fareloj's narrower variant (#917) as the base and folded in the resolved identity; your PR covered the same idea. Closing in favor of the shipped fix.

PR #896, opened 2026-05-25
feat: autoload report viewer after analysis save

urbantest28:
## Summary

- After `save_report_to_disk()` succeeds, automatically refresh the manifest, start `serve.py` on port 7788 (if not already running), and open the browser to the saved report URL
- Entire block is gated on `Trading Reports.html` existing — silent no-op in environments without the viewer
- Uses `sys.executable` instead of hardcoded `"python"` to work correctly in virtual environments
- Browser only opens when a server is confirmed listening (`_port_open(7788)`)

## Changes

- `cli/main.py` — added `import sys/socket/subprocess/webbrowser`, `_port_open()` helper, and three-step autoload block in `run_analysis()`
- `tests/test_autoload_viewer.py` — 6 unit tests for `_port_open` (TDD)

## Test Plan

- [ ] Run `pytest tests/ -m unit` — 100 tests pass
- [ ] Run an analysis end-to-end, choose Y to save — confirm browser opens to `http://localhost:7788/Trading%20Reports.html#/r/<dirname>`
- [ ] Run a second analysis — confirm only one `serve.py` process (port check prevents duplicate)
- [ ] Temporarily remove `Trading Reports.html` — confirm no error and no browser open


cheaptubes-ship-it:
how the fuck do i get out of this group. I shouldn't have to unsub from
each and every fucking sender. the spam is non stop

PR #900, opened 2026-05-26
feat(cli): add env var detection to skip interactive selections

seifenk:
## Summary
- Add env var detection for Output Language, LLM Provider, and Thinking Agents steps
- When user sets TRADINGAGENTS_* env variables in .env, CLI automatically skips corresponding interactive selection steps
- Fix: import tradingagents to trigger load_dotenv() to load .env file
- Fix: use env var value for selected_llm_provider instead of DEFAULT_CONFIG
- Fix: remove duplicate import os in get_user_selections()

## Test plan
- [ ] Verify CLI skips interactive steps when env vars are set
- [ ] Verify CLI still works normally when env vars are not set
- [ ] Verify API key check still works when using env vars

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -1,4 +1,5 @@
 from typing import Optional
+import os
 import datetime
 import typer
 import questionary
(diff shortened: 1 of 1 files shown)


Yijia-Xiao:
Thanks @seifenk — your proposal drove this. The config-layer override landed upstream via the TRADINGAGENTS_* overrides in default_config, and the CLI prompt-skip you outlined shipped in 2e67782, built on that mechanism so the env values flow through correctly (provider, models, and custom backend URL). Closing as implemented — really appreciate the contribution.

PR #902, opened 2026-05-26
fix: respect LLM env vars in CLI setup

### LINKED ISSUE #897: Unable to connect with custom .env variables
I'm using opencode go which is supposed to be compatible with the openai api format.  These are my variables (minus my API key).  However, when i run tradingagents, it still brings up the prompt to choose my LLM provider and models, which doesn't work because opencode go isn't an option. Any help would be appreciated, thanks

TRADINGAGENTS_LLM_PROVIDER=openai
TRADINGAGENTS_DEEP_THINK_LLM=kimi-k2.5
TRADINGAGENTS_QUICK_THINK_LLM=deepseek-v4-pro
TRADINGAGENTS_LLM_BACKEND_URL=https://opencode.ai/zen/go/v1

AyushKar2005:
Fixes #897

This PR fixes the CLI flow so it respects the `TRADINGAGENTS_*` LLM environment variables before showing the interactive provider/model prompts.

The problem was that even after setting custom env vars like:

```env
TRADINGAGENTS_LLM_PROVIDER=openai
TRADINGAGENTS_DEEP_THINK_LLM=kimi-k2.5
TRADINGAGENTS_QUICK_THINK_LLM=deepseek-v4-pro
TRADINGAGENTS_LLM_BACKEND_URL=https://opencode.ai/zen/go/v1
```

the CLI was still opening the prompt to choose the LLM provider and models manually.

That makes it hard to use OpenAI-compatible providers which are not listed in the dropdown, for example opencode go or any similar custom OpenAI-compatible endpoint. The env vars were already the right way to configure this, but the interactive CLI flow was not fully respecting them before asking the user again.

### What changed

I updated the CLI utility logic so it checks the env vars first.

Now:

- `TRADINGAGENTS_LLM_PROVIDER` is used before showing the provider dropdown
- `TRADINGAGENTS_LLM_BACKEND_URL` is used as the backend URL when present
- `TRADINGAGENTS_QUICK_THINK_LLM` is used before showing the quick-thinking model prompt
- `TRADINGAGENTS_DEEP_THINK_LLM` is used before showing the deep-thinking model prompt
- the project `.env` file is loaded before these CLI checks happen
- the normal questionary prompts are still used as fallback when env vars are not configured

So this does not remove the current interactive setup. It only skips the provider/model prompts when the user already configured them using env vars.

### Why this helps

This helps users who want to run TradingAgents with OpenAI-compatible APIs that are not built-in dropdown options.

For example, if someone is using a custom provider/proxy but it speaks OpenAI API format, they can now configure it through:

```env
TRADINGAGENTS_LLM_PROVIDER=openai
TRADINGAGENTS_LLM_BACKEND_URL=<custom-openai-compatible-url>
TRADINGAGENTS_QUICK_THINK_LLM=<custom-quick-model>
TRADINGAGENTS_DEEP_THINK_LLM=<custom-deep-model>
```

and the CLI will not force them back into the interactive model/provider selection.

This also keeps the experience same for normal users who dont set these env vars. They will still get the usual prompts.

### Tests added

I added tests for the new env-var behavior.

The tests check that:

- `select_llm_provider()` returns the provider and backend URL from env vars
- `_select_model(..., "quick")` returns `TRADINGAGENTS_QUICK_THINK_LLM`
- `_select_model(..., "deep")` returns `TRADINGAGENTS_DEEP_THINK_LLM`
- `questionary.select()` is not called when those env vars are already configured

This should prevent the same regression later, because the tests fail if the CLI starts prompting again even though env vars are set.

### Validation

I ran:

```bash
python -m compileall cli/utils.py
python -m pytest tests/test_cli_llm_env_vars.py -q
```

Result:

```bash
3 passed
```

### DIFF
--- a/cli/utils.py
+++ b/cli/utils.py
@@ -1,9 +1,9 @@
 import os
 from pathlib import Path
-from typing import List, Optional, Tuple, Dict
+from typing import List, Optional, Tuple
 
 import questionary
-from dotenv import find_dotenv, set_key
+from dotenv import find_dotenv, load_dotenv, set_key
 from rich.console import Console
 
 from cli.models import AnalystType, AssetType
@@ -11,6 +11,8 @@ from tradingagents.llm_clients.api_key_env import get_api_key_env
 from tradingagents.llm_clients.model_catalog import get_model_options
 
 console = Console()
+
+_DOTENV_LOADED = False
 
 TICKER_INPUT_EXAMPLES = "Examples: SPY, CNC.TO, 7203.T, 0700.HK"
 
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @AyushKar2005. CLI env-var precedence is handled now — TRADINGAGENTS_* overrides are honored and the interactive prompts skip when set. Closing as superseded.

PR #923, opened 2026-05-29
fix: make report save prompt reliable

### LINKED ISSUE #914: Save report? [Y]: y
No report file has been saved, no file location prompt, no error, nothing. 

wsl/ubuntu

Image

AyushKar2005:
Fixes #914

This PR fixes the report saving flow after the analysis is completed.

The issue was that when user typed `y` for:

```text
Save report? [Y]: y
```

nothing was really happening from the user side. No report file was saved, no file location prompt was shown, and also there was no error message. So it was hard to understand if the CLI ignored the input or if saving failed silently.

I made the save prompt handling more explicit and testable instead of doing the yes/no check directly inside `run_analysis()`.

### What changed

I added small helper functions for the post-analysis prompt flow:

- `should_save_report()`
- `should_display_report()`
- `resolve_report_save_path()`

These helpers make the prompt behaviour easier to understand and also easier to test.

Now the save report prompt properly handles normal user inputs like:

- `y`
- `Y`
- `yes`
- `YES`
- empty Enter/default value
- values with extra spaces like ` YES `

And it rejects:

- `n`
- `no`
- random/unrelated inputs

I also updated the save path logic so if the user just presses Enter for the default path, the default path is used properly. If the user gives a custom path, that path is used instead.

### Why this helps

This should make the CLI report saving flow more reliable after analysis finishes.

Before this, the prompt handling was inline and not covered by tests, so a small prompt/input edge case could make the save step confusing or look like it was silently skipped.

With this change, the logic is seperated into small helpers and tested directly.

### Tests added

I added tests for:

- accepting yes values for saving report
- rejecting no/random values
- using same yes/no rules for displaying report
- resolving default save path when input is empty
- resolving custom save path
- actually writing the complete report and section files to disk

The test also verifies that files like these are created:

```text
complete_report.md
1_analysts/market.md
2_research/manager.md
3_trading/trader.md
4_risk/neutral.md
5_portfolio/decision.md
```

### Validation

I ran:

```bash
python -m compileall cli/main.py
python -m pytest tests/test_cli_report_saving.py -q
```

Result:

```bash
6 passed, 1 warning
```

The warning is from an existing invalid escape sequence in `cli/utils.py`, not from this PR change.

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -974,6 +974,45 @@ def format_tool_args(args, max_length=80) -> str:
         return result[:max_length - 3] + "..."
     return result
 
+
+def should_save_report(value: str | None) -> bool:
+    """Return True when the user wants to save the report."""
+    if value is None:
+        return True
+
+    normalized = str(value).strip().lower()
+
+    if normalized == "":
+        return True
+
+    return normalized in {"y", "yes"}
+
+
+def should_display_report(value: str | None) -> bool:
+    """Return True when the user wants to display the report."""
+    if value is None:
+        return True
+
+    normalized = str(value).strip().lower()
+
+    if normalized == "":
+        return True
+
+    return normalized in {"y", "yes"}
+
+
+def resolve_report_save_path(value: str | None, default_path: Path) -> Path:
+    """Resolve the final report save path from prompt input."""
+    if value is None:
+        return default_path
+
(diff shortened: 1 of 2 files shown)


Tauric-Research:

Issue #862, opened 2026-05-19
Reddit API Error: HTTP Error 403: Blocked


proofconcept39:
Reddit API is closed. The work around is provided by Artistic-State: Use api this get latest https://www.reddit.com/r/webscraping/new.json
Then:
Get post detail with this: https://www.reddit.com/r/webscraping/comments/1t080rn/how_to_scrape_reddit_now_closed_api.json

Source: https://www.reddit.com/r/webscraping/comments/1t080rn/how_to_scrape_reddit_now_closed_api/

Thank you for updating! Really appreciate you. 
Max

Yijia-Xiao:
Fixed in 3543e53. When the JSON search endpoint 403s, the fetcher falls back to Reddit's public RSS search feed, which still serves anonymous requests. Fallback posts are marked "via RSS feed" since RSS lacks score/comment counts. Verified live against a blocked environment.

PR #927, opened 2026-05-30
fix(dataflows): implement Atom/RSS fallback for Reddit search (403 Blocked)

mazar:
### Problem
Reddit's security systems and WAF aggressively block anonymous public JSON search endpoints (`/r/{sub}/search.json`), returning an `HTTP Error 403: Blocked` error. This breaks the social media sentiment analyst's Reddit fetching capability out-of-the-box for all users.

### Solution
This PR adds a fallback mechanism that queries Reddit's public Atom/RSS search endpoint (`/r/{sub}/search.rss`) when the JSON endpoint fails. The RSS/Atom feed is much more open and accepts requests using browser-like User-Agents.

### Technical Details
- Catches HTTP/network errors on the JSON request and transparently invokes `_fetch_subreddit_rss()`.
- Parses the namespaced Atom XML feed using the built-in `xml.etree.ElementTree` parser.
- Extracts and sanitizes post body excerpts from the HTML nested inside the `<content>` tag.
- Converts Atom timestamp strings to UTC timestamps.
- Returns a list of dictionaries identical in structure to the original JSON parser schema (defaulting engagement statistics like `score` and `num_comments` to `0` since they are not exposed in the RSS feed), maintaining 100% backward compatibility.

### DIFF
--- a/tradingagents/dataflows/reddit.py
+++ b/tradingagents/dataflows/reddit.py
@@ -12,9 +12,13 @@ never have to special-case missing data.
 
 from __future__ import annotations
 
+import html
 import json
 import logging
+import re
 import time
+import xml.etree.ElementTree as ET
+from datetime import datetime
 from typing import Iterable
 from urllib.error import HTTPError, URLError
 from urllib.parse import urlencode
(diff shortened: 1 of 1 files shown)


mazar:
May fix #862

Yijia-Xiao:
Thanks @mazar — shipped in 3543e53 on your RSS-fallback approach. We dropped the pinned browser User-Agent (testing showed the RSS feed serves our normal app UA fine) and marked RSS-sourced posts so they don't show fake zero scores. Closing in favor of the shipped fix — appreciate the clean diagnosis.

PR #930, opened 2026-05-31
Add default configuration memory

### LINKED ISSUE #920: Feature Request: Add default configuration memory to avoid repeated manual selection
Currently, when using the interactive Trading Agents project, users need to manually select a large number of configuration options every time they start the program. All selections need to be repeated manually for each run, which brings redundant and repetitive operation costs to daily use.
### Expected Feature
I hope to add a local default configuration memory function to optimize user interaction experience:
1. Automatically save all user configuration options after the first complete manual selection;
2. When restarting the program next time, provide two optional entry modes directly: Use last default configuration or Re-select parameters manually;
3. Support overwriting the default configuration freely: users can update the saved default value after finishing a new round of manual selection.

AyushKar2005:
## What this PR does

This PR adds a small local config memory for the interactive CLI setup.

Right now, whenever users start the TradingAgents CLI, they have to manually select the same options again and again like ticker, analysts, research depth, provider, models, output language etc. This gets repetative for daily usage.

With this change, after the user completes the setup once, the selected configuration is saved locally in:

```bash
~/.tradingagents/default_config.json
```

Next time when the CLI starts, it checks if this saved config exists and asks the user if they want to reuse it or configure everything manually again.

## Changes made

* Added local default config save/load logic
* Added startup prompt when saved config exists
* Added option to use saved configuration directly
* Added option to configure manually again
* Manual configuration overwrites the old saved config
* Converted selected analysts into JSON safe values before saving
* Updated analysis flow to support saved analyst values also

## Why this is useful

This should reduce repeated manual selections for users who mostly run the CLI with same settings. It keeps the existing manual flow same, but gives users a faster path after first setup.

This is mostly focused on improving CLI user experience and fixing the repeated config selection issue.

## Testing done

I tested it manually with the CLI.

First I removed old saved config:

```bash
rm -f ~/.tradingagents/default_config.json
```

Then I ran the CLI:

```bash
python -m cli.main
```

I completed the setup manually once and verified that config file was created:

```bash
cat ~/.tradingagents/default_config.json
```

Then I ran the CLI again:

```bash
python -m cli.main
```

It showed the saved config prompt and I was able to choose:

```txt
Use saved configuration
```

Also checked the file compiles:

```bash
python -m compileall cli/main.py
```

## Notes

The config is stored locally in user's home folder, so it does not touch project files or git tracked files.

Fixes #920

Screenshot 2026-05-31 193106

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -1,5 +1,6 @@
 from typing import Optional
 import datetime
+import json
 import typer
 import questionary
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


Tauric-Research:

PR #931, opened 2026-05-31
fix(anthropic): skip effort kwarg when quick model does not support it

tcusolle:
Haiku and Sonnet reject the `effort` parameter with a 400 error; only Opus supports it.

This guards the kwarg so mixed quick/deep model configs (e.g. Sonnet quick + Opus deep) do not break on startup with a 400 from the Anthropic API.

**Root cause:** `effort` kwargs were being forwarded to both quick and deep clients unconditionally. The fix checks that both models contain `"opus"` before passing the parameter.

**Reproduction:** Set `quick_think_llm` to any Sonnet or Haiku model with `anthropic_effort` set — the run fails immediately with a 400 error before any agent executes.


Yijia-Xiao:
Thanks @tcusolle. The Anthropic effort parameter is gated to models that accept it (Opus 4.5+ / Sonnet 4.6+; Haiku excluded). Closing as superseded.

PR #938, opened 2026-06-01
GUI Development stage 01

lanraeee:


praxstack:
_Automated principal-engineer review (generated by Claude Code). Findings below were adversarially verified; please treat them as a starting point for discussion rather than blocking decrees._

Thanks for the substantial work here — the Anthropic temperature-gating fix and its tests are clean, and the web layer is a nice direction. A few things stand out before this is safe to ship, grouped by severity.

## Critical

### Unauthenticated live-money order execution behind wildcard CORS
**File:** `web/app.py` (execute endpoint ~lines 637-687), `web/broker.py` (~lines 755-758)

**Problem:** `POST /api/analyze/{analysis_id}/execute` calls `broker.place_order` with no authentication, no session/CSRF check, and no real confirmation token — the "user-confirmed" docstring is only a UI affordance. The server-side guards are limited to analysis-exists / proposed_order-present / not-already-executed. Compounding this: the broker defaults to **live** trading (`os.getenv("ALPACA_PAPER", "false")` → falsy default → `paper=False` → live base URL), and CORS is `allow_origins=["*"]` together with `allow_credentials`.

**Failure scenario:** Any origin (or anyone who can reach the host) can POST to the execute endpoint and place real orders against a live Alpaca account, with no credential ever checked. A malicious page in the user's browser can drive it via the credentialed wildcard CORS config.

**Suggested fix:**
- Require authentication on all `/api` routes (at minimum a server-side API token via a `Depends` guard).
- Default `ALPACA_PAPER` to paper/true; require an explicit opt-in for live trading.
- Replace `allow_origins=["*"]` with an explicit allowlist, and either drop `allow_credentials` or pair it with concrete origins (the two are mutually exclusive in the CORS spec anyway).
- Issue a per-analysis confirmation token to the initiating session and require it on execute.

## Medium

### Auto-selecting a different provider silently ignores the user's model choice
**File:** `web/app.py` (`start_analysis` / `_resolve_provider_models`, ~lines 77-78, 317-341)

**Problem:** In `start_analysis`, when the requested provider has no key, `_first_configured_provider()` may return `minimax`, `minimax-cn`, or `openrouter` — none of which are in `PROVIDER_DEFAULT_MODELS`. `_resolve_provider_models` then returns early, leaving the previously-requested provider's model name paired with the newly-selected provider.

**Failure scenario:** User requests OpenAI with a `gpt-5.5` model but has no OpenAI key; the code falls back to `openrouter` while keeping `gpt-5.5` as the model name — producing exactly the provider/model mismatch the function exists to prevent, and a confusing downstream failure.

**Suggested fix:** When auto-selecting, realign the models too — either add catalog defaults for `openrouter`/`minimax`, or reject auto-selection for providers without known defaults and return a clear `400` asking the user to pick a model.

## Low

### `active_analyses` mutated from a sync background thread while async handlers read it
**File:** `web/app.py` (`run_analysis_task` and the websocket/GET readers)

**Problem:** `run_analysis_task` runs in a thread (a `def` function under `BackgroundTasks`) and mutates `active_analyses[id]` status/progress/messages, while the websocket loop and GET handlers read the same dict from the event loop. Individual dict/list ops are GIL-atomic, so state won't corrupt, but readers can observe torn intermediate states, and there's no concurrency bound beyond `_evict_old_analyses`.

**Failure scenario:** Acceptable for a single-user demo, but fragile for the multi-user service this PR positions — concurrent runs can show inconsistent progress, and eviction could drop a non-terminal analysis.

**Suggested fix:** Guard the store with a lock (or an asyncio-friendly structure), and never evict a non-terminal analysis — cap concurrent runs instead.

### New top-level report/changelog `.md` files add deployment/marketing noise to the repo root
**File:** `VERIFICATION_REPORT.md` (and `RAILWAY_FIX.md`, `WEB_APP_SUMMARY.md`, `DEPLOYMENT_RAILWAY.md`, `WEB_QUICKSTART.md`, `web/README.md`)

**Problem:** Seven new files (~1800 lines) read as session changelogs rather than durable docs, and hardcode fork-specific details (`fawazzzbello/strattonoak`, a personal feature branch). `VERIFICATION_REPORT.md` asserts "ALL SYSTEMS GO / Zero Errors," which is contradicted by the security default above — an unverified green claim.

**Failure scenario:** Future readers trust the "Zero Errors" report and ship the live-trading default; the fork-specific references rot once merged upstream.

**Suggested fix:** Drop the fix/verification/summary changelog files; keep a single concise `web/README.md` referencing the upstream repo with no hardcoded fork branch.

## Overall recommendation

Request changes. The Anthropic fix is mergeable on its own merits, but the web layer should not land until the live-money execute endpoint is authenticated, defaults to paper trading, and the CORS config is tightened — these are the blockers. The provider/model realignment is a worthwhile correctness fix, and the concurrency and docs items are good cleanups for the multi-user framing. Happy to help review the follow-ups.

🤖 Automated review via Claude Code

praxstack:
_Proposed implementation of the fixes from my earlier automated review, offered as a ready-to-apply patch. Generated by an automated principal-engineer pass; please treat as a starting point._

**What it does:** Auth-gate all `/api` routes (optional `TRADINGAGENTS_API_TOKEN`, constant-time compare), remove the credentialed wildcard CORS, default the Alpaca broker to **PAPER** (live = explicit opt-in), add a per-analysis confirmation token required on `/execute`, fix a swallowed-HTTPException bug, and remove ~1500 lines of fork-specific changelog files. Adds `tests/test_web_security.py` (25 tests).

The patch is **2,310 lines** — too large for a GitHub comment, so it's here as a gist:

➡️ **- Creating gist pr-938.patch
✓ Created secret gist pr-938.patch
https://gist.github.com/praxstack/0135f2f07ab5fd4fd99a620ffd83a818**

Apply it on top of this PR's branch (the diff is against this PR's head commit):

```bash
curl -L https://gist.githubusercontent.com/praxstack/0135f2f07ab5fd4fd99a620ffd83a818/raw/pr-938.patch -o pr-938.patch
git apply pr-938.patch
```

The biggest item is the **critical** one: `POST /api/analyze/{id}/execute` currently places real-money orders with no auth and live-by-default; the patch closes that. Happy to walk through any of it.

🤖 Automated patch via Claude Code — not pushed anywhere; yours to apply, adapt, or ignore.

praxstack:
@fawazzzbello Opened a PR **into your branch** with the security fix (auth on all `/api` routes, paper-trading default, tightened CORS, execute confirmation token; +25 tests): fawazzzbello/StrattonOak#3 — merge, adapt, or cherry-pick. 🤖

PR #952, opened 2026-06-02
Add Chinese A-share news (East Money) and fix non-US ticker robustness

akaponym-byte:
## Summary

Adds Chinese-language news coverage for A-shares and fixes three robustness/UX issues that surface when analyzing non-US tickers. Discovered while running the framework on Shanghai/Shenzhen tickers (e.g. `600519.SS`, `300450.SZ`).

## Motivation

When analyzing an A-share, the English-centric sources have little to no coverage, and two of them actively broke the run:

- The Sentiment Analyst crashed the **entire** run with an unhandled `UnicodeEncodeError` when a non-ASCII symbol reached the StockTwits request line.
- The News Analyst had no real news for A-shares (Yahoo/Alpha Vantage return next to nothing for `.SS`/`.SZ`).
- Library `logger.warning` calls (StockTwits 404 / Reddit 403) wrote straight to stderr and tore the Rich `Live` dashboard.
- A live layout taller than the terminal tiled the header banner down the screen on every refresh.

## Changes (one focused commit each)

1. **fix(dataflows): stop non-ASCII tickers from crashing StockTwits fetch** — catch `UnicodeError` so the fetcher degrades gracefully like every other failure it already handles, instead of crashing the run.
2. **feat(dataflows): add East Money Chinese news for A-shares with auto-routing** — new `get_news_eastmoney` vendor (东方财富, no API key) mirroring `get_news_yfinance`'s signature/format. Registered as a `get_news` vendor and auto-preferred for `.SS`/`.SZ` tickers. Market-aware routing is required because `get_news_yfinance` returns an empty-but-successful string for A-shares, which would otherwise shadow any fallback. Other markets are unaffected; an explicit vendor config still wins. Adds unit tests.
3. **fix(cli): route library logs to a file so they don't corrupt the Live display** — attach a root file handler at the start of `run_analysis` (logs to `results_dir/tradingagents.log`), which also stops logging's lastResort handler from writing to stderr during the Live render.
4. **fix(cli): render progress dashboard on the alternate screen** — pass `screen=True` to `Live` so frames repaint in place instead of tiling. The full report is still printed to the normal screen after the Live block.

## Testing

- New `tests/test_eastmoney_news.py` (8 tests): A-share detection, JSON/JSONP parsing + date filtering + tag stripping, graceful degradation on HTTP error, and the routing preference (A-share → East Money, US → yfinance).
- Full suite green: `319 passed, 75 subtests passed`.
- Manually verified live fetches return real Chinese news for `600519.SS`, `000858.SZ`, and `603688.SS`, and that US tickers (`AAPL`) still route to yfinance.

## Notes

- East Money is a `get_news` (per-stock) vendor only; `get_global_news` and the social sources are unchanged.
- No new dependencies; the source uses the stdlib HTTP client and needs no API key.

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -1,6 +1,7 @@
 from typing import Optional
 import os
 import datetime
+import logging
 import typer
 import questionary
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


cheaptubes-ship-it:
fucking spammers

Tauric-Research:

PR #963, opened 2026-06-04
Handle invalid News Analyst tool calls without aborting

0x-genesys:
## Summary
- Retry the News Analyst once when the model emits malformed tool-call arguments.
- If the retry still produces invalid tool calls, log the issue and continue the agent loop instead of crashing the run.
- Normal news-analysis tool usage is unchanged; this only hardens the failure path.

## Validation
- `python -m py_compile tradingagents/agents/analysts/news_analyst.py`

## Notes
- This addresses the DeepSeek 400 caused by a malformed `get_news` tool call in the news analyst turn.

### DIFF
--- a/tradingagents/agents/analysts/news_analyst.py
+++ b/tradingagents/agents/analysts/news_analyst.py
@@ -1,3 +1,6 @@
+import logging
+
+from langchain_core.messages import AIMessage, HumanMessage, ToolMessage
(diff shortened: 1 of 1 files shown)


0x-genesys:
Closing to reopen as a fresh PR for a clean review pass after updating the tool-call retry logic.

PR #995, opened 2026-06-05
feat(llm): add custom OpenAI-compatible provider

thieucong98:


Yijia-Xiao:
Thanks @thieucong98 — same goal. Implemented in 20d3b07 as a first-class `openai_compatible` provider, now part of a unified provider registry, with tests and a real end-to-end check. Closing as addressed — appreciate the contribution.

PR #996, opened 2026-06-06
fix: declare python-dotenv dependency

### LINKED ISSUE #994: Bug: python-dotenv is imported by the CLI but not declared as a direct dependency
## Summary
The CLI imports `dotenv.find_dotenv` and `dotenv.set_key`, but `python-dotenv` is not listed in `pyproject.toml` dependencies. A clean install may fail at import time depending on transitive dependencies.

## Evidence
- `cli/utils.py:6` imports from `dotenv`.
- `pyproject.toml:7-29` dependency list does not include `python-dotenv`.

## Reproduction
Install the project in a clean environment where no transitive package provides python-dotenv, then run `tradingagents`.

## Expected
All directly imported packages should be direct dependencies.

## Suggested fix
Add `python-dotenv>=1.0` to `pyproject.toml`.

AyushKar2005:
Fixes #994

This PR adds `python-dotenv` as a direct runtime dependency.

The CLI imports from `dotenv` in `cli/utils.py`, specifically `find_dotenv`, `load_dotenv`, and `set_key`. But `python-dotenv` was not declared directly in `pyproject.toml`.

Because of that, a clean install could fail at runtime with:

```bash
ModuleNotFoundError: No module named 'dotenv'
```

if no transitive package happens to install it.

### What changed

- Added `python-dotenv>=1.0` to `pyproject.toml`
- Verified the CLI still compiles
- Verified the dotenv imports work after installing the package

### Why this helps

Since the CLI directly imports `dotenv`, it should also declare `python-dotenv` directly instead of relying on some other dependency to bring it indirectly.

This makes clean installs more reliable.

### Validation

I ran:

```bash
python -m compileall cli
python -m pip install -e .
python -c "from dotenv import find_dotenv, set_key; print('python-dotenv import ok')"
```

Result:

```bash
python-dotenv import ok
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,7 @@ dependencies = [
     "pandas>=2.3.0",
     "parsel>=1.10.0",
     "pytz>=2025.2",
+    "python-dotenv>=1.0",
     "questionary>=2.1.0",
     "redis>=6.2.0",
     "requests>=2.32.4",


Yijia-Xiao:
Thanks @AyushKar2005. python-dotenv is already declared on main (`2a58c22`); closing as resolved.

PR #1001, opened 2026-06-07
fix(cli): label quick vs deep model prompts for OpenRouter

### LINKED ISSUE #1000: BUG (CLI): OpenRouter model prompts are indistinguishable for Quick-Thinking vs Deep-Thinking
## Description
When OpenRouter is selected as the LLM provider, the CLI presents two model-selection steps with the identical prompt text: `Select OpenRouter Model (latest available)`. There is nothing in either prompt indicating that the first selection sets the quick-thinking model (quick_think_llm) and the second sets the deep-thinking model (deep_think_llm). The roles are distinguishable only by call order, which isn't surfaced to the user.

## Steps to reproduce
1. Run tradingagents (CLI).
2. Select provider: OpenRouter.
3. Proceed to the Thinking Agents step.
4. Observe the two consecutive model-selection prompts.
Image

## Expected behavior
Each prompt should state its role, consistent with how other providers (e.g. OpenAI) label them as "Quick-Thinking" / "Deep-Thinking" — e.g.:

`Select Your [Quick-Thinking] OpenRouter Model (latest available):`
`Select Your [Deep-Thinking] OpenRouter Model (latest available):`

yting27:
Fix #1000 by giving each prompt a model-specific label consistent with the existing provider flows.

### DIFF
--- a/cli/utils.py
+++ b/cli/utils.py
@@ -206,20 +206,22 @@ def select_openrouter_model(mode: str) -> str:
     ).ask()
 
     if choice is None or choice == "custom":
-        return questionary.text(
+        custom_id = questionary.text(
             "Enter OpenRouter model ID (e.g. google/gemma-4-26b-a4b-it):",
             validate=lambda x: len(x.strip()) > 0 or "Please enter a model ID.",
-        ).ask().strip()
+        ).ask()
+        return custom_id.strip() if custom_id else ""
 
     return choice
 
 
 def _prompt_custom_model_id() -> str:
     """Prompt user to type a custom model ID."""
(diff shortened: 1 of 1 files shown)


Yijia-Xiao:
Thanks @yting27 — shipped in c15200dc286b66abce3f1bcf09b298dc06b8539d, reimplemented against current main (your branch conflicted with the full-repo lint cleanup). Kept the cancel-safety idea and extended it across the other prompts, and the OpenRouter dropdown now shortlists the newest models from mainstream providers. Appreciate the report and the fix.

Issue #992, opened 2026-06-05
Bug: Flat yfinance news articles bypass historical date filtering


RSXLX:
## Summary
For flat yfinance news article structures, `_extract_article_data()` returns `pub_date=None`. The caller only filters by date when `pub_date` exists, so historical analyses can include current/out-of-window news.

## Evidence
- `tradingagents/dataflows/yfinance_news.py:44-51` flat article fallback sets `pub_date` to `None`.
- `tradingagents/dataflows/yfinance_news.py:88-92` skips date filtering when `pub_date` is missing.

## Expected
Flat article timestamps should be parsed where available, or undated articles should be handled conservatively for historical runs.

## Actual
Out-of-window news can enter the prompt.

## Suggested fix
Parse common flat timestamp fields (`providerPublishTime`, `publishTime`, etc.) and decide whether to exclude undated articles in historical mode.

Yijia-Xiao:
Fixed in 0c1231a. yfinance news now parses the flat-format publish time and applies one look-ahead-safe window filter across ticker and global news, so future-dated (and, in a historical/backtest window, undated) articles no longer leak in. Verified with real data and regression tests. Thanks @RSXLX.

Issue #993, opened 2026-06-05
Bug: Global news can return an empty report after filtering future articles


RSXLX:
## Summary
`get_global_news_yfinance()` slices `all_news[:limit]` before filtering out articles published after the requested date. If those first articles are filtered, the function may return an empty news section without a no-news message, even though later collected articles might be usable.

## Evidence
- `tradingagents/dataflows/yfinance_news.py:173` iterates `all_news[:limit]`.
- `tradingagents/dataflows/yfinance_news.py:177-181` can skip items after the slice.
- `tradingagents/dataflows/yfinance_news.py:199` returns the header regardless of whether `news_str` is empty.

## Expected
Filtering should happen before applying the final limit, and empty post-filter results should return a clear no-news message.

## Suggested fix
Build a filtered list first, then slice to `limit`, and check whether the formatted body is empty.

Yijia-Xiao:
Fixed in 0c1231a: when every candidate falls outside the window, global news now returns an explicit "No global news found between ..." message instead of an empty-bodied report. Thanks @RSXLX.

PR #1008, opened 2026-06-08
fix(yfinance): improve article publication date extraction and limit handling

yting27:
To fix issues #992 and #993.
Changes made:
- Fallback to parse publish time field in older API (`providerPublishTime`).
- Skip news article if `pub_date` is empty.
- Filter article date before applying limit.


Yijia-Xiao:
Thanks @yting27 — same idea. Implemented in 0c1231a: flat articles now parse providerPublishTime so they're date-filterable, with regression tests. Closing as addressed — appreciate the contribution.

PR #1015, opened 2026-06-09
Respect explicit single-vendor routing config

### LINKED ISSUE #988: Bug: Explicit single-vendor data config still silently falls back to other vendors
## Summary
The router always appends remaining vendors to the fallback chain. This means a config like `core_stock_apis: alpha_vantage` does not actually mean Alpha Vantage only.

## Evidence
- `tradingagents/dataflows/interface.py:138-149` starts with configured vendors, then appends all other available vendors.

## Expected
A single configured vendor should be authoritative, or there should be a separate explicit fallback setting.

## Actual
The app silently uses other vendors after primary failure/rate-limit/no-data.

## Suggested fix
Only fallback when the config explicitly lists multiple vendors, e.g. `alpha_vantage,yfinance`, or add a `fallback_enabled` option.

gyx09212214-prog:
## Summary
- make a single configured data vendor authoritative instead of appending every other vendor automatically
- keep fallback available when config explicitly lists multiple vendors, e.g. `alpha_vantage,yfinance`
- preserve `default`/empty config behavior by routing through all vendors
- add router regression tests for single-vendor no-fallback and explicit comma-separated fallback

Fixes #988.

## Tests
- `.\.venv\Scripts\python -m pytest tests\test_no_data_handling.py tests\test_dataflows_config.py -q`
- `git diff --check`

### DIFF
--- a/tests/test_no_data_handling.py
+++ b/tests/test_no_data_handling.py
@@ -15,6 +15,7 @@ import pandas as pd
 import pytest
 
 from tradingagents.dataflows import stockstats_utils, interface
+from tradingagents.dataflows.alpha_vantage_common import AlphaVantageRateLimitError
 from tradingagents.dataflows.config import set_config
 from tradingagents.dataflows.symbol_utils import NoMarketDataError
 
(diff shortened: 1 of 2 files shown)


gyx09212214-prog:
Automated review follow-up pushed in `215a846`.

Handled Gemini feedback:
- `_vendor_chain` now accepts list-style vendor config without stringifying brackets/quotes into vendor names.
- `default` can be mixed into an explicit vendor chain and expands to the remaining configured vendors without duplicating earlier entries.
- Added regression coverage for list config and mixed `default` fallback routing.

Validation:
- `python -m py_compile tradingagents/dataflows/interface.py tests/test_no_data_handling.py`: passed
- `git diff --check`: passed
- `python -m pytest tests/test_no_data_handling.py -q`: 7 passed

Yijia-Xiao:
Thanks @gyx09212214-prog — same fix. Implemented in 6560883: the configured vendor list is now the exact chain (no silent fallback to unconfigured vendors), with regression tests. Closing as addressed — appreciate the contribution.

PR #1016, opened 2026-06-09
Fix OHLCV cache download horizon

### LINKED ISSUE #985: Bug: OHLCV cache comment says 15 years but implementation downloads only 5 years
## Summary
The OHLCV loader documents a 15-year cache window, but the implementation only downloads 5 years. Historical analyses older than five years can lose technical indicator data unexpectedly.

## Evidence
- `tradingagents/dataflows/stockstats_utils.py:68` says downloads 15 years.
- `tradingagents/dataflows/stockstats_utils.py:83` uses `pd.DateOffset(years=5)`.

## Expected
The implementation should match the documented window or the documentation should be corrected.

## Impact
Backtests or historical analysis dates older than five years may produce empty/no-data indicator snapshots even when Yahoo has data.

## Suggested fix
Use `years=15` or update the comment and user-facing expectations to 5 years.

gyx09212214-prog:
## Summary
- use the documented 15-year OHLCV cache horizon when downloading yfinance history
- keep the cache-window value in one constant so the documentation and implementation stay aligned
- add a regression test that verifies `load_ohlcv` requests roughly 15 years of data without touching the network

Fixes #985.

## Tests
- `python -m pytest tests\test_no_data_handling.py -q`
- `git diff --check -- tradingagents/dataflows/stockstats_utils.py tests/test_no_data_handling.py`

### DIFF
--- a/tests/test_no_data_handling.py
+++ b/tests/test_no_data_handling.py
@@ -45,6 +45,30 @@ class TestLoadOhlcvNoPoison(unittest.TestCase):
                 stockstats_utils.load_ohlcv("FAKE", "2026-01-01")
             self.assertTrue(dl2.called)
 
+    def test_download_window_matches_documented_cache_horizon(self):
+        downloaded = pd.DataFrame(
+            {
+                "Open": [100.0],
+                "High": [101.0],
+                "Low": [99.0],
+                "Close": [100.5],
+                "Volume": [1000],
+            },
+            index=pd.DatetimeIndex(["2025-01-02"], name="Date"),
+        )
+
+        with mock.patch.object(stockstats_utils.yf, "download", return_value=downloaded) as dl:
+            stockstats_utils.load_ohlcv("AAPL", "2025-12-31")
+
+        kwargs = dl.call_args.kwargs
+        start = pd.Timestamp(kwargs["start"])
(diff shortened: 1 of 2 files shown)


gyx09212214-prog:
Addressed the Gemini cache review in 62f1bc5.

Changes:
- Switched OHLCV cache files to a stable per-symbol/per-horizon name instead of embedding daily start/end dates.
- Reuse a cache when it covers the requested date, or when it has already been refreshed today.
- Remove older date-named OHLCV cache files for the same symbol after a successful refresh.
- Added regression coverage for cross-day cache reuse and legacy cache cleanup.

Validation:
- python -m pytest tests/test_no_data_handling.py -q (5 passed)
- git diff --check -- tradingagents/dataflows/stockstats_utils.py tests/test_no_data_handling.py
- AST syntax check for the two touched files

Yijia-Xiao:
Thanks @gyx09212214-prog. We resolved the doc/impl mismatch by keeping the 5-year window and correcting the comment (in dab0768) rather than expanding to 15 years: 5 years already covers every indicator we compute (the 200-day SMA needs ~1 year), so the extra history would 3x the download for no analytical gain. Closing — appreciate flagging it.

PR #1017, opened 2026-06-09
Honor env backend URL after provider selection

### LINKED ISSUE #978: Bug: TRADINGAGENTS_LLM_BACKEND_URL is ignored unless TRADINGAGENTS_LLM_PROVIDER is also set
## Summary
A custom backend URL from the environment is only used in the env-provider path. If the user chooses a provider interactively, the selected provider's default URL overwrites `TRADINGAGENTS_LLM_BACKEND_URL`.

## Evidence
- `tradingagents/default_config.py:14` maps `TRADINGAGENTS_LLM_BACKEND_URL`.
- `cli/main.py:571-578` uses it only when `TRADINGAGENTS_LLM_PROVIDER` is set.
- `cli/main.py:585` otherwise uses `select_llm_provider()`'s URL.

## Reproduction
Set `TRADINGAGENTS_LLM_BACKEND_URL` to a proxy or LM Studio endpoint, do not set `TRADINGAGENTS_LLM_PROVIDER`, then pick the provider interactively.

## Expected
The env backend URL should be honored independently of provider selection.

## Suggested fix
After provider selection, prefer `DEFAULT_CONFIG["backend_url"]` when non-empty.

gyx09212214-prog:
## Summary
- honor `TRADINGAGENTS_LLM_BACKEND_URL` after an interactive provider selection
- keep provider-specific regional selection intact, but let the explicit env backend URL override the provider default endpoint
- add regression coverage for the case where only the backend URL is set in the environment and the provider is still selected interactively

Fixes #978.

## Tests
- `python -m pytest tests\test_cli_env_skip.py -q`
- `git diff --check -- cli/main.py tests/test_cli_env_skip.py`


chenguangyu (commented):
Thanks for fixing #978 — the behavioral change is clear and the new test covers the right scenario.

**What looks good**
- After interactive provider (and regional) selection, re-applying `DEFAULT_CONFIG["backend_url"]` matches the env-only provider path and fixes the override gap.
- `test_backend_url_env_overrides_interactive_provider_default` correctly asserts that the env URL wins over the interactive default.

**Request before merge**
- The diff shows ~2.6k lines changed in `cli/main.py`, but with `--ignore-cr-at-eol` the real change is only ~4 lines. It looks like the file was converted to CRLF end-to-end, which makes the PR very hard to review and pollutes blame history.
- Please re-save `cli/main.py` (and `tests/test_cli_env_skip.py` if affected) with LF line endings and push a follow-up commit so the PR only contains the functional + test changes.

Once the line-ending noise is cleaned up, this should be good to merge.

gyx09212214-prog:
Addressed in fe2158a.

Changes:
- Re-saved cli/main.py and tests/test_cli_env_skip.py with LF line endings.
- PR diff is now reduced to the real changes: cli/main.py 6 additions / 2 deletions, tests/test_cli_env_skip.py 33 additions.

Validation:
- python -m pytest tests/test_cli_env_skip.py -q (5 passed)
- git diff --check -- cli/main.py tests/test_cli_env_skip.py
- confirmed both files have 0 CRLF line endings

Yijia-Xiao:
Thanks @gyx09212214-prog. An explicit env backend URL now wins regardless of how the provider was chosen (resolve_backend_url, #978). Closing as superseded.

PR #1022, opened 2026-06-11
fix: reject stale yfinance OHLCV data

### LINKED ISSUE #1021: Incorrect market data: ticker prior date close price is often incorrect in reports
Incorrect market data: last closed quote price used for analysis is same day but 1 year old

Examples:

Ticker for analysis made on June 11, 2026    | Actual market close price on June 10, 2026                 |
CB      | "closing at $284.45 on June 10"       |  $330.58                                                                            |
INCY  | "Current Price ~$70.07"                     |  $105.39                                                                            |

The discrepancy is big. And if this is taken as basis for technical analysis it makes such reports useless.

On analysis performed on June 11, 2026, a report (CB) contains the following: 
**"using data through June 11, 2025 (reflecting the most recent available trading data)"**
I.e., from time to time, it uses exactly same date, **but one year old** price quotes.

AyushKar2005:
Fixes #1021

This PR adds a stale-data guard for yfinance OHLCV data before it is used by the analyst workflow.

The issue reported cases where analysis for June 2026 was using quote data from June 2025, basically same month/day but one year old. That can make the current close price, technical indicators, and final market report badly incorrect.

### What changed

- Added a shared stale OHLCV check in `stockstats_utils.py`
- The check compares the latest available OHLCV row with the requested analysis date
- If the latest row is too far before the requested date, it raises a stale market data error
- Kept the latest upstream symbol normalization logic, including `normalize_symbol()` and `NoMarketDataError`
- Applied the stale-data guard to:
  - `load_ohlcv()` used by stockstats / indicator paths
  - `get_YFin_data_online()` used by raw yfinance stock data output
- Added handling so stale data returns a clear message instead of silently feeding old prices into the report
- Added tests for stale OHLCV behavior

### Why this helps

This prevents the workflow from silently using old market data as if it was the latest available quote.

For example, if the requested analysis date is:

```text
2026-06-11
```

but the latest OHLCV row returned by yfinance/cache is:

```text
2025-06-11
```

the data layer now rejects it and surfaces the stale-data problem instead of letting the LLM build technical analysis using the wrong close price.

### Tests added

The tests cover:

- accepting a recent prior trading day
- rejecting one-year-old stale data
- rejecting empty OHLCV data
- returning a stale-data message from `get_YFin_data_online()`

### Validation

I ran:

```bash
python -m compileall tradingagents/dataflows/stockstats_utils.py tradingagents/dataflows/y_finance.py
python -m pytest tests/test_yfinance_stale_ohlcv_guard.py -q
```

Result:

```bash
4 passed
```

### DIFF
--- a/tests/test_yfinance_stale_ohlcv_guard.py
+++ b/tests/test_yfinance_stale_ohlcv_guard.py
@@ -1,0 +1,88 @@
+from __future__ import annotations
+
+import pandas as pd
+import pytest
+
+import tradingagents.dataflows.y_finance as y_finance
+from tradingagents.dataflows.stockstats_utils import (
+    StaleMarketDataError,
+    _assert_ohlcv_not_stale,
+)
+
+
+def test_ohlcv_stale_guard_accepts_recent_prior_trading_day():
+    data = pd.DataFrame(
+        {
+            "Date": [pd.Timestamp("2026-06-10")],
+            "Open": [330.0],
+            "High": [332.0],
+            "Low": [328.0],
+            "Close": [330.58],
+            "Volume": [1000000],
+        }
+    )
+
+    _assert_ohlcv_not_stale(data, "2026-06-11", "CB")
+
+
+def test_ohlcv_stale_guard_rejects_one_year_old_same_month_day():
+    data = pd.DataFrame(
+        {
+            "Date": [pd.Timestamp("2025-06-11")],
+            "Open": [280.0],
+            "High": [286.0],
+            "Low": [278.0],
+            "Close": [284.45],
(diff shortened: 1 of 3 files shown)


Yijia-Xiao:
Thanks @AyushKar2005 — clear diagnosis and a solid guard. Shipped in 9fd54f83682284fed82ec54e3719a12fd440a0f9, refined a little: it raises the existing NoMarketDataError with a stale-specific detail (rather than a new error class), so the vendor router's try-next-vendor and single unavailable-signal path apply unchanged, and the sentinel surfaces the reason. Appreciate the fix and the tests.

PR #1026, opened 2026-06-12
fix(reddit): go RSS-first to avoid 403→429 rate-limit churn (follow-up to #862)

LukiPrince:
## Problem

The social-media analyst's Reddit fetch returns no posts on most runs: every subreddit logs `Reddit JSON fetch failed ... 403 Blocked` followed by `Reddit RSS ... 429 Too Many Requests`, so the sentiment section ends up blank.

## Root cause

`dataflows/reddit.py` tries the JSON search endpoint first and only falls back to RSS on failure. But:

- `/search.json` is **reliably** WAF-blocked (`HTTP 403`) for public / non-OAuth clients — the module docstring already documents this (#862).
- So every call pays for a doomed JSON request **and** an RSS request: ~2× the volume against Reddit's per-IP rate limit.
- The 403s return instantly, which defeats the `inter_request_delay` pacing, so a normal 3-subreddit run fires ~6 requests in ~1.6s and 429s on the 2nd/3rd RSS call.

Reproduced from a single egress IP:

| endpoint | status |
|---|---|
| `https://www.reddit.com/r/stocks/search.json?...` | **403** |
| `https://www.reddit.com/r/stocks/search.rss?...`  | **200** |

The RSS feed itself works fine; it's the doubled volume + instant-403 pacing that trips the limit.

## Fix

- **RSS-first** — the default per-subreddit fetch (`_fetch_subreddit`) now goes straight to `/search.rss`, halving request volume. The JSON path is preserved as `_fetch_subreddit_json` (still degrades to RSS on 403) for the day the WAF relaxes or an OAuth token is wired in.
- **429 backoff** — the RSS path backs off once on `429`, honouring `Retry-After` (capped at 30s), before giving up, so a transient burst no longer blanks the feed.
- Pace per-subreddit requests `0.4s → 1.0s`.

No change to the formatter or returned data shape: RSS posts are still tagged and score/comment counts still omitted when absent.

## Testing

`tests/test_reddit_fallback.py`:
- retargets the existing JSON→RSS fallback test at `_fetch_subreddit_json`,
- adds RSS-first delegation for `_fetch_subreddit` (asserts the JSON endpoint is never hit),
- adds 429 coverage: retry-once-then-success, give-up-after-one-retry, and `Retry-After` honoured.

```
13 passed
```

Also verified against a live run: the social analyst now receives Reddit posts (e.g. r/wallstreetbets) instead of an empty section.

### DIFF
--- a/tests/test_reddit_fallback.py
+++ b/tests/test_reddit_fallback.py
@@ -1,4 +1,5 @@
-"""Tests for the Reddit RSS/Atom fallback when the JSON endpoint 403s (#862)."""
+"""Tests for the RSS-first Reddit fetcher, its 429 backoff and the opt-in
+JSON path's degradation (#862)."""
 
 from __future__ import annotations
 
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @LukiPrince — excellent diagnosis of the 403→429 churn. Shipped RSS-first with the one-shot Retry-After backoff and wider pacing (3cddf1e331dbbd5d993ace62edbb58cb3dd65713 and the two commits before it). Appreciate the clear write-up and the reproduction.

PR #1029, opened 2026-06-13
fix(cli): declare python-dotenv dependency

Gujiassh:
# Summary
- declare `python-dotenv` as a direct dependency because the CLI imports it directly
- add a focused metadata regression test so direct CLI imports stay represented in `pyproject.toml`
- keep the fix scoped to packaging metadata without unrelated lockfile churn

# Testing
- `python3 -m pytest tests/test_cli_dependency_metadata.py -q`
- `python3 -m py_compile tests/test_cli_dependency_metadata.py`
- `git diff --check`

# Breaking Changes
- [x] None

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ dependencies = [
     "tqdm>=4.67.1",
     "typing-extensions>=4.14.0",
     "yfinance>=1.4.1",
+    "python-dotenv>=1.1.0",
 ]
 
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @Gujiassh. `python-dotenv` is already declared as a direct dependency on main (2a58c22), so this is in place. Closing as superseded — appreciate the diligence.

PR #1030, opened 2026-06-13
fix(deps): regenerate stale uv.lock to match pyproject; clears 41 CVEs

macd2:
## What

Regenerate the committed `uv.lock` so it is consistent with `pyproject.toml`. **No `pyproject.toml` changes** — this is a pure lockfile fix.

## Why

The committed `uv.lock` is **inconsistent with the declared dependencies**, so `uv lock --check` fails and any `uv sync` / `uv lock` silently re-resolves the dependency graph:

- `pyproject.toml` requires `langchain-google-genai>=4.0.0` (raised in `bdc5fc6` for thought-signature support).
- `langchain-google-genai>=4.0.0` hard-requires `langchain-core>=1.1`.
- But the committed lock pins `langchain-google-genai 2.1.5` and `langchain-core 0.3.x` — which violates the above and is mutually unsatisfiable.

On top of that, the stale pins carry a large number of disclosed CVEs in transitive packages (aiohttp, urllib3, lxml, idna, orjson, protobuf, pyasn1, pygments, python-dotenv, requests, marshmallow, langsmith, langgraph, etc.).

## Change

`uv lock --upgrade` — regenerate the lockfile to satisfy `pyproject.toml`. This resolves the stack to its intended majors and brings transitive deps current:

| Package | Before (lock) | After |
|---|---|---|
| langchain-core | 0.3.83 | 1.4.7 |
| langchain-openai | 0.3.23 | 1.3.2 |
| langchain-text-splitters | 0.3.8 | 1.1.2 |
| langgraph | 0.4.8 | 1.2.5 |
| langsmith | 0.3.45 | 0.8.15 |
| langchain-google-genai | 2.1.5 | 4.2.5 |
| aiohttp | 3.12.13 | 3.14.1 |
| urllib3 | 2.4.0 | 2.7.0 |
| lxml | 5.4.0 | 6.1.1 |

## Verification

- **`uv lock --check`**: now passes (lock consistent with `pyproject.toml`).
- **`pip-audit`**: before — **41 known CVEs across 11 packages**; after — **no known vulnerabilities found**.
- **Test suite**: `312 passed, 1 skipped (live DeepSeek API call, no key), 75 subtests passed`. The code runs cleanly on the resolved LangChain 1.x stack — confirming the intended majors were already code-compatible and only the lockfile lagged.

## Notes

- No source or `pyproject.toml` changes — only `uv.lock`.
- This is the minimal change that makes the repo installable as declared; happy to split the LangChain major bump out if maintainers prefer to stage it, but the test suite is green as-is.


Yijia-Xiao:
Thanks @macd2. Addressed in `0b61eff`: the committed `uv.lock` is removed — it isn't used by the pip-based install or CI and only drifts. The CVE count came from the stale lock's own pins; the pyproject dependencies are unaffected. A deliberate LangChain upgrade would be its own scoped PR.

PR #1033, opened 2026-06-14
fix(dataflows): log swallowed vendor errors

nyxst4ck:
## Summary
- log swallowed exceptions when a primary data vendor fails and routing falls back
- add regression coverage for the vendor routing failure path

## Why
Without this, a broken primary vendor can be silently masked by fallback behavior, making production diagnosis harder.

## Test Plan
- pytest tests/test_route_vendor_logging.py

### DIFF
--- a/tests/test_route_vendor_logging.py
+++ b/tests/test_route_vendor_logging.py
@@ -20,14 +20,14 @@ class TestRouteToVendorLogging(unittest.TestCase):
             return "FALLBACK_DATA"
 
         patched = {"boomvendor": primary_boom, "okvendor": fallback_ok}
-        with mock.patch.object(interface, "get_vendor", return_value="boomvendor,okvendor"), \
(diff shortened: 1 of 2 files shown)


what168gp-jpg (approved):

nyxst4ck:
Updated after review: removed the redundant exception argument from the exc_info=True warning and added explicit coverage/logging for configured-primary vendor rate-limit fallback. Validation: python -m pytest tests/test_route_vendor_logging.py -q; python -m ruff check tradingagents/dataflows/interface.py tests/test_route_vendor_logging.py.

Yijia-Xiao:
Thanks @nyxst4ck. The vendor router already logs swallowed vendor failures on main (6560883, building on your earlier #1010), so a broken primary is no longer silent. Closing as superseded — thanks for the follow-up.

PR #1034, opened 2026-06-14
Add default token/LLM-call budgets + ignore .DS_Store

muralisastry:
## Summary
- `Add default token and LLM-call budgets` — seed the same per-run spend caps in `tradingagents/default_config.py` that the TradingAgents-Core webapp enforces, so CLI/programmatic entry points read the same defaults. (Commit was already local but unpushed.)
- Ignore `.DS_Store`.

## Verification
`python -m pytest` (base conda env) → **108 passed, 42 subtests passed**.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -217,3 +217,6 @@ __marimo__/
 
 # Cache
 **/data_cache/
+
+# macOS
+.DS_Store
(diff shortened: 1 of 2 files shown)


muralisastry:
Opened against upstream by mistake — intended for my own fork. Closing.

PR #1035, opened 2026-06-14
fix(fred): validate series_id and degrade gracefully on invalid LLM indicators

Handsomemikezzz:
## Summary
Hardens the FRED macro vendor against invalid `series_id` values passed by LLM tool calls to `get_macro_indicators`.
When the News Analyst paraphrases an indicator (e.g. a full English sentence instead of a short alias like `cpi` or `unemployment`), the previous resolver uppercased the raw string and sent it to FRED as `series_id`. FRED rejects IDs that are not ≤25 alphanumeric characters, which raised `ValueError` and crashed the entire research graph because `macro_data` has no fallback vendor.
This PR:
- validates `series_id` before calling FRED
- adds fuzzy alias resolution for common LLM paraphrases
- returns an `INVALID_FRED_INDICATOR` tool message instead of raising for bad indicators / unknown series
- keeps `FredNotConfiguredError` and non-series `ValueError` behavior unchanged
## Problem / reproduction
1. Configure `macro_data: fred` with a valid `FRED_API_KEY`
2. Run a full graph where the News Analyst calls `get_macro_indicators`
3. If the LLM passes a long free-text indicator, FRED returns:
   `Bad Request. Invalid value for variable series_id. Series IDs should be 25 or less alphanumeric characters.`
4. `route_to_vendor("get_macro_indicators", ...)` re-raises the error (single vendor, no fallback), failing the whole task
## Test plan
- [x] `pytest tests/test_fred.py -q` (16 passed)
- [x] invalid / overlong indicators return `INVALID_FRED_INDICATOR` without calling FRED
- [x] router surfaces invalid indicators without crashing

### DIFF
--- a/tests/test_fred.py
+++ b/tests/test_fred.py
@@ -99,11 +99,29 @@ class FredFormattingTests(unittest.TestCase):
             out = fred.get_macro_data("unemployment", "2025-09-30", 30)
         self.assertIn("No observations", out)
 
-    def test_unknown_series_raises(self):
+    def test_unknown_series_returns_invalid_indicator_message(self):
         no_series = {"seriess": []}
-        with mock.patch.object(fred, "_request", side_effect=_request_stub(meta=no_series)), \
-                self.assertRaises(ValueError):
-            fred.get_macro_data("totally_unknown_xyz", "2025-09-30", 30)
+        with mock.patch.object(fred, "_request", side_effect=_request_stub(meta=no_series)):
+            out = fred.get_macro_data("totally_unknown_xyz", "2025-09-30", 30)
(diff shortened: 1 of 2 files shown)


what168gp-jpg (approved):

Handsomemikezzz:
Addressed in f79e76f — generic aliases are now deprioritized; added the suggested test cases.

Yijia-Xiao:
Thanks @Handsomemikezzz for the graceful-FRED report. Same fix shipped in `ee1ece3` (alongside #1054): macro data degrades to a sentinel instead of crashing.

PR #1036, opened 2026-06-14
fix(dataflows): data-source resilience — circuit breaker, off-switch, graceful degradation

Eddieargenal:
## Summary

Three independent fixes so non-critical data vendors can't spam logs, waste retries, or abort an expensive run. All surfaced running the full pipeline where Polymarket's API was unreachable and an optional macro key was absent.

### 1. Polymarket circuit breaker (`dataflows/polymarket.py`)
`get_prediction_markets` is called once per analyst topic. When the Gamma host is hard-unreachable (DNS / connection refused), the old code re-attempted and re-logged for **every** topic. Now a **hard** failure (`requests.ConnectionError`) trips a per-process breaker — logged once, subsequent topics short-circuit — while **transient** failures (timeout / HTTP) keep retrying.

### 2. "Off" switch in `route_to_vendor` (`dataflows/interface.py`)
Setting a category to `""`/`none`/`off`/`disabled` previously built an empty vendor chain and raised `ValueError`. It now returns a graceful "disabled in configuration" sentinel, letting users opt out of a flaky/irrelevant source. Example: `config["data_vendors"]["prediction_markets"] = "off"`.

### 3. Graceful degradation of optional sources (`dataflows/interface.py`)
`route_to_vendor` raised `first_error` when no vendor could serve a method. For an **enrichment** category with a single vendor (`macro_data` → fred, `prediction_markets` → polymarket), a missing `FRED_API_KEY` or unreachable host propagated up and **aborted the whole LangGraph run** — dozens of paid LLM calls deep. Now `OPTIONAL_CATEGORIES = {macro_data, prediction_markets}` degrade to an instructive sentinel (mirroring the existing `NO_DATA_AVAILABLE` path) so the analyst proceeds without that signal. Core categories (price/indicators/fundamentals) still raise, so a broken primary stays loud.

## Tests
- `tests/test_polymarket_resilience.py` (9) — breaker trips on hard failure & skips repeats; transient failures don't trip it; success path intact; disabled category routes to sentinel without invoking the vendor.
- `tests/test_optional_source_degradation.py` (4) — optional sources degrade; core categories still raise; the existing NO_DATA path is preserved.
- Updates the FRED router test to the new contract (degrade, don't raise).

`ruff check` clean; related vendor/routing suites green.

### DIFF
--- a/tests/test_fred.py
+++ b/tests/test_fred.py
@@ -157,9 +157,12 @@ class FredRoutingTests(unittest.TestCase):
             out = interface.route_to_vendor("get_macro_indicators", "cpi", "2026-06-01", 365)
         self.assertEqual(out, "MACRO_OK")
 
-    def test_not_configured_surfaces_through_router(self):
-        # With only fred and no key, the router has no fallback and must surface
-        # the real "not configured" failure rather than masking it.
+    def test_not_configured_degrades_gracefully_through_router(self):
+        # macro_data is an OPTIONAL (enrichment) category: with only fred and no
+        # key, the router must NOT abort the run — a missing optional key
+        # destroying an expensive, near-complete run is the worst failure mode.
+        # Instead it returns an instructive sentinel so the analyst proceeds
+        # without the macro signal (see interface.OPTIONAL_CATEGORIES).
         set_config({"data_vendors": {"macro_data": "fred"}})
 
         def _unconfigured(*a, **k):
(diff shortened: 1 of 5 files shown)


Yijia-Xiao:
Thanks @Eddieargenal for the data-resilience work. Shipped the optional-vendor degradation at the router layer in `ee1ece3` (covers FRED and Polymarket generally). Left out the process-global Polymarket breaker: it would stick across runs in the long-lived web worker, and Polymarket already degrades per call.

PR #1037, opened 2026-06-14
feat: preflight validation, report parity, usage/cost accounting, run manifest

Eddieargenal:
## Summary

Four related improvements to run setup, output, and observability. Each is independent and behind a config flag (all default-on except strict preflight). Surfaced while running the full pipeline programmatically (the Python API produced far thinner artifacts than the CLI, and a run's cost was invisible).

### #2 — Preflight config validation (`tradingagents/config_validation.py`)
Maps the selected analysts → data categories → required API keys and checks them at `TradingAgentsGraph` init, instead of discovering a missing key deep inside a node after minutes of paid calls. Missing **optional** sources (macro / prediction-markets) only warn; missing **required** ones warn and, with `config["preflight_strict"]`, raise before any LLM call.

### #3 — Report parity (`tradingagents/reporting.py`)
Extracts the markdown report-tree writer out of `cli/main.py` into a reusable library function. The CLI now delegates to it (no behavior change), and `TradingAgentsGraph` writes the same tree after a run (`config["write_reports"]`, on by default). Programmatic callers previously got only a single JSON blob.

### #5 — Usage / cost accounting (`tradingagents/usage.py`)
A provider-agnostic `UsageTrackingCallback` aggregates LLM/tool call counts and token usage (from each provider's `usage_metadata`), with an optional dollar estimate when `config["model_prices"]` is set — no built-in price table that would silently go stale. Attached to every run by default; a one-line summary is logged at the end. The CLI's `StatsCallbackHandler` now subclasses it, making the library the single source of truth.

### #6 — Run manifest (`tradingagents/manifest.py`)
Writes `manifest.json` beside the results: model/provider, **API-key presence as booleans only (never the values)**, data-vendor coverage, analysts, the preflight result, and the usage tally. Makes runs auditable and comparable.

## Config flags (all opt-out except strict preflight)
- `preflight_strict` (default `false`) — raise on a missing required source
- `write_reports` (default `true`) — write the markdown report tree from the API
- `track_usage` (default `true`) — attach the usage callback
- `model_prices` (default `None`) — `{"input": usd_per_1M, "output": usd_per_1M}` for cost estimates
- `write_manifest` (default `true`) — write manifest.json

## Tests
New: `test_preflight_validation`, `test_reporting`, `test_usage_tracking`, `test_manifest`. Full suite: **470 passed** (4 pre-existing, unrelated failures in temperature/ollama client tests are not touched by this PR). `ruff check` clean.

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -48,6 +48,7 @@ from tradingagents.graph.analyst_execution import (
     sync_analyst_tracker_from_chunk,
 )
 from tradingagents.graph.trading_graph import TradingAgentsGraph
+from tradingagents.reporting import write_reports
 
 console = Console()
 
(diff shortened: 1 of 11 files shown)


Yijia-Xiao:
Thanks @Eddieargenal for the thorough bundle. Shipped the report-parity piece in `a0120e1`: a shared writer so programmatic runs save the same report tree the CLI does (`TradingAgentsGraph.save_reports`). Left out the usage tracker (overlaps the shipped StatsCallbackHandler) and the run manifest (depends on it) — happy to revisit those separately if there's demand.

PR #1039, opened 2026-06-15
fix(docker): use TRADINGAGENTS_LLM_PROVIDER for ollama profile

### LINKED ISSUE #975: Bug: Docker Ollama profile sets LLM_PROVIDER instead of TRADINGAGENTS_LLM_PROVIDER
## Summary
The docker-compose Ollama profile sets `LLM_PROVIDER=ollama`, but the application config only reads `TRADINGAGENTS_LLM_PROVIDER`. As a result, `docker compose --profile ollama run --rm tradingagents-ollama` can still start with the default OpenAI provider instead of Ollama.

## Evidence
- `docker-compose.yml:23` sets `LLM_PROVIDER=ollama`.
- `tradingagents/default_config.py:10-20` only maps `TRADINGAGENTS_LLM_PROVIDER` to `llm_provider`.

## Expected
The Ollama compose service should configure the provider as Ollama.

## Actual
The provider remains whatever `DEFAULT_CONFIG` resolves to unless the user also sets `TRADINGAGENTS_LLM_PROVIDER=ollama`.

## Suggested fix
Change the compose env var to `TRADINGAGENTS_LLM_PROVIDER=ollama` and set `OLLAMA_BASE_URL` to the compose service URL if needed.

ydhawesome:
## Summary
The `tradingagents-ollama` Compose service sets `LLM_PROVIDER=ollama`, but config overrides are only read from `TRADINGAGENTS_*` env vars (`default_config._ENV_OVERRIDES`). The bare `LLM_PROVIDER` is never read, so the profile silently falls back to the default provider instead of Ollama.

This also adds `OLLAMA_BASE_URL` pointing at the sibling container's service name. With the provider correctly switched to ollama, the backend URL resolves (via `resolve_backend_url` → `provider_default_url`) to `OLLAMA_BASE_URL or http://localhost:11434/v1`. Inside the app container `localhost` resolves to itself, not the `ollama` service, so without this the profile cannot reach the model server.

## Changes
```yaml
    environment:
      - TRADINGAGENTS_LLM_PROVIDER=ollama
      - OLLAMA_BASE_URL=http://ollama:11434/v1
```

## Testing
- `docker compose config` parses; YAML validated.
- Confirmed `_ENV_OVERRIDES` maps only `TRADINGAGENTS_LLM_PROVIDER`, and the Ollama base URL is resolved from `OLLAMA_BASE_URL` in both the provider table and `resolve_backend_url`.

Fixes #975

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,7 +20,12 @@ services:
     env_file:
       - .env
     environment:
-      - LLM_PROVIDER=ollama
+      # Config overrides are read from TRADINGAGENTS_* env vars (see
+      # default_config._ENV_OVERRIDES); a bare LLM_PROVIDER is ignored.
+      - TRADINGAGENTS_LLM_PROVIDER=ollama
+      # Reach the sibling ollama container by service name, not localhost,
+      # which inside this container would resolve to itself.
+      - OLLAMA_BASE_URL=http://ollama:11434/v1
     volumes:
       - tradingagents_data:/home/appuser/.tradingagents
     depends_on:


Yijia-Xiao:
Thanks @ydhawesome — spot-on diagnosis. Shipped in a420ad0, implemented on current main with `TRADINGAGENTS_LLM_PROVIDER` plus `OLLAMA_BASE_URL` for the sibling ollama service. Closing in favor of the merged fix, credited to you.

PR #1040, opened 2026-06-15
fix(cli): preserve env debate round overrides

### LINKED ISSUE #977: Bug: CLI interactive research depth overrides TRADINGAGENTS_MAX_DEBATE_ROUNDS
## Summary
Environment-configured debate rounds are overwritten by the interactive research-depth selection. This makes `TRADINGAGENTS_MAX_DEBATE_ROUNDS` and `TRADINGAGENTS_MAX_RISK_ROUNDS` ineffective for normal CLI runs.

## Evidence
- `tradingagents/default_config.py:16-17` maps both env vars.
- `cli/main.py:997-998` assigns both values from `selections["research_depth"]`.

## Expected
If debate-round env vars are set, the CLI should preserve them or skip the research-depth prompt.

## Actual
The prompt value always wins.

## Suggested fix
When either env var is present, skip or decouple the research-depth override for that specific config key.

Ghraven:
## Problem

`TRADINGAGENTS_MAX_DEBATE_ROUNDS` and `TRADINGAGENTS_MAX_RISK_ROUNDS` are applied into `DEFAULT_CONFIG`, but the interactive CLI then overwrites both values with the selected research depth before constructing the graph. That means a user can set the env vars and still get the menu-selected round count instead.

Closes #977

## Before / after

Before:
- `DEFAULT_CONFIG` could contain env-configured debate/risk round values.
- `run_analysis()` always replaced both values with `selections["research_depth"]`.

After:
- The interactive research depth still applies when no round env vars are set.
- Explicit `TRADINGAGENTS_MAX_DEBATE_ROUNDS` / `TRADINGAGENTS_MAX_RISK_ROUNDS` values are preserved through CLI config assembly.

## Verification

- `python -m pytest tests/test_cli_env_skip.py tests/test_env_overrides.py`
- `python -m ruff check cli/main.py tests/test_cli_env_skip.py tests/test_env_overrides.py`
- `git diff --check`

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -1019,14 +1019,22 @@ def format_tool_args(args, max_length=80) -> str:
         return result[:max_length - 3] + "..."
     return result
 
+
+def _apply_research_depth(config: dict, research_depth: int) -> None:
+    """Apply interactive research depth unless env vars already configured it."""
+    if not os.environ.get("TRADINGAGENTS_MAX_DEBATE_ROUNDS"):
+        config["max_debate_rounds"] = research_depth
+    if not os.environ.get("TRADINGAGENTS_MAX_RISK_ROUNDS"):
+        config["max_risk_discuss_rounds"] = research_depth
+
+
 def run_analysis(checkpoint: bool = False):
     # First get all user selections
     selections = get_user_selections()
 
     # Create config with selected research depth
     config = DEFAULT_CONFIG.copy()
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @Ghraven. Shipped in a420ad0: the env round counts now win over the interactive research-depth selection (the prompt is skipped when both are set). Implemented on current main; closing in favor of the merged fix, credited to you.

PR #1041, opened 2026-06-15
fix(cli): preserve env checkpoint setting

### LINKED ISSUE #976: Bug: CLI ignores TRADINGAGENTS_CHECKPOINT_ENABLED when --checkpoint is omitted
## Summary
`TRADINGAGENTS_CHECKPOINT_ENABLED=true` is applied to `DEFAULT_CONFIG`, but the CLI overwrites it with the Typer option default `False` unless the user explicitly passes `--checkpoint`.

## Evidence
- `tradingagents/default_config.py:18` maps `TRADINGAGENTS_CHECKPOINT_ENABLED`.
- `cli/main.py:1008` assigns `config["checkpoint_enabled"] = checkpoint`.
- `cli/main.py:1290-1294` defaults `checkpoint` to `False`.

## Reproduction
Set `TRADINGAGENTS_CHECKPOINT_ENABLED=true` in `.env`, run `tradingagents analyze` without `--checkpoint`, and observe that checkpointing is disabled.

## Expected
The CLI should honor env config unless the command-line flag explicitly overrides it.

## Suggested fix
Use `DEFAULT_CONFIG["checkpoint_enabled"]` as the base and only override when the CLI option is explicitly set.

Ghraven:
## Problem

`TRADINGAGENTS_CHECKPOINT_ENABLED` is applied to `DEFAULT_CONFIG`, but the CLI command default passes `checkpoint=False` into `run_analysis()` whenever `--checkpoint` is omitted. That overwrites an env-enabled checkpoint setting before the graph is constructed.

Closes #976

## Before / after

Before:
- `TRADINGAGENTS_CHECKPOINT_ENABLED=true` was reflected in `DEFAULT_CONFIG`.
- Running `tradingagents analyze` without `--checkpoint` still set `checkpoint_enabled` back to `False`.

After:
- Omitting the checkpoint flag preserves the env/default config value.
- `--checkpoint` explicitly enables checkpointing.
- `--no-checkpoint` explicitly disables checkpointing when a user wants to override an env-enabled default.

## Verification

- `python -m pytest tests/test_cli_env_skip.py tests/test_env_overrides.py`
- `python -m ruff check cli/main.py tests/test_cli_env_skip.py tests/test_env_overrides.py`
- `git diff --check`

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -1019,7 +1019,13 @@ def format_tool_args(args, max_length=80) -> str:
         return result[:max_length - 3] + "..."
     return result
 
-def run_analysis(checkpoint: bool = False):
+
+def _apply_checkpoint_config(config: dict, checkpoint: bool | None) -> None:
+    if checkpoint is not None:
+        config["checkpoint_enabled"] = checkpoint
+
+
+def run_analysis(checkpoint: bool | None = None):
     # First get all user selections
     selections = get_user_selections()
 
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @Ghraven. Shipped in a420ad0: `--checkpoint` is now tri-state so omitting it preserves `TRADINGAGENTS_CHECKPOINT_ENABLED`. Implemented on current main; closing in favor of the merged fix, credited to you.

Issue #984, opened 2026-06-05
Bug: Memory reflection return lookup bypasses symbol normalization


RSXLX:
## Summary
The deferred memory reflection path fetches post-decision returns with the raw ticker, not the canonical Yahoo symbol. Instruments accepted through alias normalization may never resolve outcomes.

## Evidence
- `tradingagents/graph/trading_graph.py:240` calls `yf.Ticker(ticker).history(...)`.
- Market data tools normalize aliases in `tradingagents/dataflows/symbol_utils.py`.

## Reproduction
Run an analysis for `BTCUSD` or `XAUUSD+`, then run the same ticker again after enough holding days.

## Expected
Pending memory entries should resolve using the same canonical symbol as market data.

## Actual
The raw ticker may return no yfinance rows, leaving the memory entry pending.

## Suggested fix
Normalize the ticker in `_fetch_returns()` before calling yfinance.

Yijia-Xiao:
Fixed in 7c8fe2f. `normalize_symbol` is now applied on this path too, so broker/forex/commodity symbols resolve to the same Yahoo symbol the price path uses (e.g. `XAUUSD`/`GOLD` -> `GC=F`, `BTCUSD` -> `BTC-USD`). Verified with real calls (identity now returns Gold/Bitcoin/EUR-USD instead of empty) and unit tests guard the regression. Thanks for the precise report @RSXLX.

Issue #983, opened 2026-06-05
Bug: Instrument identity lookup bypasses symbol normalization


RSXLX:
## Summary
Instrument identity resolution queries yfinance with the raw uppercase ticker instead of the normalized Yahoo symbol, so aliases that work for market data can fail identity resolution.

## Evidence
- `tradingagents/agents/utils/agent_utils.py:75` calls `yf.Ticker(ticker.upper()).info`.
- `tradingagents/dataflows/symbol_utils.py` maps aliases like `BTCUSD -> BTC-USD`, `XAUUSD+ -> GC=F`, `US500 -> ^GSPC`.

## Reproduction
Analyze `BTCUSD`, `XAUUSD+`, or `US500`.

## Expected
Identity context should resolve against the canonical Yahoo symbol.

## Actual
Identity lookup can return `{}`, causing weaker prompt anchoring.

## Suggested fix
Call `normalize_symbol(ticker)` before the yfinance identity request.

Yijia-Xiao:
Fixed in 7c8fe2f. `normalize_symbol` is now applied on this path too, so broker/forex/commodity symbols resolve to the same Yahoo symbol the price path uses (e.g. `XAUUSD`/`GOLD` -> `GC=F`, `BTCUSD` -> `BTC-USD`). Verified with real calls (identity now returns Gold/Bitcoin/EUR-USD instead of empty) and unit tests guard the regression. Thanks for the precise report @RSXLX.

Issue #980, opened 2026-06-05
Bug: CLI ticker validation rejects supported Yahoo symbols such as GC=F and XAUUSD+


RSXLX:
## Summary
The CLI rejects characters that the data layer explicitly supports for Yahoo/futures/broker symbols.

## Evidence
- `cli/utils.py:38` only allows `._-^`.
- `tradingagents/dataflows/symbol_utils.py:11-15` documents `GC=F` and `XAUUSD+`.
- `tradingagents/dataflows/utils.py:10-15` allows `=` and `+` for safe path components.

## Reproduction
Run the CLI and enter `GC=F` or `XAUUSD+`.

## Expected
The CLI should accept symbols supported by normalization and data fetch code.

## Actual
The prompt validation rejects them.

## Suggested fix
Allow `=` and `+` in `get_ticker()` validation.

Yijia-Xiao:
Fixed in 76add90. The CLI now resolves tickers through the same `normalize_symbol` the data path uses, so validation accepts `=` symbols (`GC=F`, `EURUSD=X`), `BTCUSD` classifies as crypto, and `BTC-USDT` normalizes to the priceable `BTC-USD` pair (verified with real data + 25 regression tests). Thanks @RSXLX.

Issue #781, opened 2026-05-09
Incorrect price analysis for XAUUSD+


tedix-rodrigo:
Hey guys,

I've just started playing around with the tradingagents via docker.  My first analysis was on "FMG" the price reads correct. However, for gold "XAUUSD+", it giving me a price of "431.08" current market value is at "4,715.72". 

Any ideas on what's going on here? I'm using the following setting: 

Model: google gemini 3 flash
Language: English
Analyst team: Market
Reach depth: Shallow

<img width="1684" height="327" alt="Image" src="https://github.com/user-attachments/assets/8a1799c8-7239-4771-b46a-6c55d654d9d2" />

Yijia-Xiao:
Fixed in 2a7ad54. TradingAgents now resolves common commodity/forex/crypto symbols to the instrument Yahoo Finance serves, so both `XAUUSD+` and `GC=F` work for gold (likewise `USOIL`->WTI, `EURUSD`, `BTCUSD`, `SPX500`, etc.). Unknown symbols now report data as unavailable rather than returning an estimated price.

Thanks @tedix-rodrigo for the report. Feel free to reopen if the issue persists.

Yijia-Xiao:
Correction: the fix landed as 1ff3f07 on main. The earlier-referenced commit was superseded during a rebase onto current main.

PR #1045, opened 2026-06-15
fix(data): normalize ticker news symbols

gyx09212214-prog:
## Summary
- accept a trailing broker CFD `+` suffix in CLI ticker validation while still rejecting misplaced or standalone `+` characters
- normalize ticker-specific yfinance news lookups through `normalize_symbol`, matching the market data, identity, and reflection paths
- surface the resolved Yahoo symbol in ticker-news output when it differs from the cleaned requested broker symbol, e.g. `GC=F (from XAUUSD+)`
- add regression coverage for the remaining `XAUUSD+` CLI boundary and ticker-news lookup path

## Why this matters
Broker-style commodity symbols such as `XAUUSD+` are already supported by the data-layer normalizer and safe ticker handling, but the CLI still rejected the trailing `+` form and ticker-specific yfinance news still queried the raw symbol directly. That left one symbol-provenance path inconsistent with the rest of the pipeline.

Keeping ticker-news queries on the canonical Yahoo symbol reduces the chance that sentiment/news context silently goes missing or diverges from the market data path for broker, forex, commodity, and crypto aliases.

## Scope
This keeps the change limited to ticker input validation and yfinance ticker-news lookup normalization. It does not add new symbol aliases, change global/macro news search, or touch graph execution/prompt behavior.

This complements the existing symbol-normalization fixes around #781, #980, #983, and #984 rather than reopening those broader paths.

## Validation
- `python -m pytest tests/test_cli_symbol_handling.py tests/test_symbol_normalization_paths.py tests/test_symbol_utils.py tests/test_news_lookahead.py tests/test_no_data_handling.py -q` - 52 passed
- `python -m pytest -q` - 448 passed, 2 skipped, 74 subtests passed
- `python -m ruff check .` - clean
- `git diff --check` - clean

### DIFF
--- a/cli/utils.py
+++ b/cli/utils.py
@@ -27,11 +27,16 @@ def is_valid_ticker_input(value: str) -> bool:
     """Whether a ticker entry is acceptable (charset + length).
 
     Allows the characters Yahoo symbols use, including ``=`` for futures/forex
-    like ``GC=F`` and ``EURUSD=X`` (#980), and ``^`` for indices. Empty input is
-    allowed (it defaults to SPY downstream).
+    like ``GC=F`` and ``EURUSD=X`` (#980), ``+`` for broker CFD symbols like
+    ``XAUUSD+``, and ``^`` for indices. Empty input is allowed (it defaults to
+    SPY downstream).
     """
     v = value.strip()
-    return not v or (all(ch.isalnum() or ch in "._-^=" for ch in v) and len(v) <= 32)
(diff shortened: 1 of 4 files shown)


Yijia-Xiao:
Thanks @gyx09212214-prog for the news-symbol fix. Shipped in `9ad98c5`: the news path now normalizes the ticker (XAUUSD to GC=F) like the price and identity paths, with provenance shown.

PR #1051, opened 2026-06-16
fix(structured): guard against None return from with_structured_output

LukiPrince:
## Problem

When a thinking-mode model (e.g. `deepseek-v4-flash`, `deepseek-reasoner`) receives a tool-calling request without `tool_choice: required` — which is intentional, since these models reject that parameter — it can decide to respond with plain text instead of invoking the tool. LangChain's `PydanticToolsParser` then **silently returns `None`** rather than raising an exception.

This caused `render(result)` to be called with `result = None`, producing a confusing and hard-to-diagnose error:

```
Portfolio Manager: structured-output invocation failed
('NoneType' object has no attribute 'rating'); retrying once as free text
```

The `AttributeError` points at `render_pm_decision` and looks like a schema bug, when the real cause is the upstream `None` return.

## Fix

Add an explicit `None` check in `invoke_structured_or_freetext` before calling `render()`:

```python
result = structured_llm.invoke(prompt)
if result is None:
    raise ValueError(
        "model skipped tool call — structured output parser returned None"
    )
return render(result)
```

The `except Exception` block already in place catches this `ValueError` and triggers the free-text fallback exactly as before. The pipeline is not blocked. The log message is now actionable:

```
Portfolio Manager: structured-output invocation failed
(model skipped tool call — structured output parser returned None); retrying once as free text
```

## Test

Added `test_falls_back_to_freetext_when_structured_returns_none` to `TestSentimentAnalystAgent` — mocks `structured_llm.invoke` to return `None` and asserts the free-text fallback fires correctly.

All 21 existing unit tests continue to pass.

### DIFF
--- a/tests/test_structured_agents.py
+++ b/tests/test_structured_agents.py
@@ -27,6 +27,7 @@ from tradingagents.agents.schemas import (
 )
 from tradingagents.agents.trader.trader import create_trader
 
+
 # ---------------------------------------------------------------------------
 # Render functions
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @LukiPrince for the None-guard. Shipped in `517eeaf`: a structured call that returns no parsed result (a thinking model answering in plain text) now falls back to free text with a clear reason instead of an opaque render error.

PR #1054, opened 2026-06-16
fix(fred): validate series IDs and fail fast on descriptive names

Dado-hash:
## Problem
When an LLM or user passed a descriptive indicator name like `bank of japan rate` to the FRED macro tool, the code treated it as a raw FRED series ID. FRED rejected it with a 400 error, producing an opaque failure.

## Changes
- Validate resolved FRED series IDs against FRED's rule: 1-25 alphanumeric characters.
- Return a clear, actionable error message from `get_macro_data()` instead of propagating a FRED 400.
- Expand `MACRO_SERIES` with additional common aliases (`headline_cpi`, `gdp_growth`, `nfp`, `dxy`, etc.).
- Update the tool docstring to list supported aliases and warn against descriptive names.
- Add a regression test for descriptive/invalid indicator names.

## Testing
- `pytest tests/test_fred.py -q` → 14 passed
- `ruff check tests/test_fred.py tradingagents/dataflows/fred.py tradingagents/agents/utils/macro_data_tools.py` → all checks passed

### DIFF
--- a/tests/test_fred.py
+++ b/tests/test_fred.py
@@ -103,7 +103,16 @@ class FredFormattingTests(unittest.TestCase):
         no_series = {"seriess": []}
         with mock.patch.object(fred, "_request", side_effect=_request_stub(meta=no_series)), \
                 self.assertRaises(ValueError):
-            fred.get_macro_data("totally_unknown_xyz", "2025-09-30", 30)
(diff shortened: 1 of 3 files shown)


Yijia-Xiao:
Thanks @Dado-hash for the FRED series-id validation. Shipped in `ee1ece3`: invalid indicators return guidance instead of 400ing, and the router now degrades optional macro/prediction data rather than aborting the run.

PR #1058, opened 2026-06-17
Fix trader proposal float parsing 3638388716509737140

Rudytheredhead:


Yijia-Xiao:
Thanks @Rudytheredhead for flagging the trader float-parsing issue. Fixed in `0405168`: the optional price fields coerce null-ish strings ('None'/'N/A') to None, so a placeholder no longer fails validation. Implemented as a focused schema validator; the PR's bundled workspace files weren't applicable.

PR #1061, opened 2026-06-19
fix: suppress local compatible tool choice objects

### LINKED ISSUE #1057: Local OpenAI-compatible servers (LM Studio) reject object-form tool_choice in structured output — not covered by capability table
## Summary

Local OpenAI-compatible servers (LM Studio, and likely llama.cpp / older vLLM) reject the **object form** of `tool_choice` that LangChain's structured-output path sends by default. Because these models aren't in the per-model capability table, they fall through to `_DEFAULT` (`supports_tool_choice=True`), so every structured-output call 400s and drops to the free-text fallback.

This is the same class of bug already fixed for DeepSeek in #678 / #689 and for MiniMax — the capability-table pattern is in place, local providers just aren't covered by it.

## Error

```
<Agent>: structured-output invocation failed
(Error code: 400 - {'error': "Invalid tool_choice type: 'object'. Supported string values: none, auto, required"});
retrying once as free text
```

Seen on the Sentiment Analyst, Trader, and Portfolio Manager stages — every agent that uses `with_structured_output`.

## Reproduce

1. Run LM Studio (Developer tab → server on) with any chat model loaded, e.g. `qwen3-coder-30b`.
2. Configure for the generic local provider:
   ```
   TRADINGAGENTS_LLM_PROVIDER=openai_compatible
   TRADINGAGENTS_LLM_BACKEND_URL=http://localhost:1234/v1
   TRADINGAGENTS_DEEP_THINK_LLM=qwen3-coder-30b
   TRADINGAGENTS_QUICK_THINK_LLM=qwen3-coder-30b
   ```
3. Run a propagation (`python main.py` / `tradingagents`). Every structured-output agent logs the 400 above and falls back to free text.

The run still completes (the fallback works), but structured output never succeeds, so agents rely on brittle free-text parsing instead of typed results.

## Root cause

`tradingagents/llm_clients/capabilities.py` resolves capabilities by model ID. LM Studio model IDs aren't listed, so `get_capabilities()` returns `_DEFAULT` with `supports_tool_choice=True`. `NormalizedChatOpenAI.with_structured_output` (`openai_client.py`) then lets LangChain send the function-spec object as `tool_choice`, which the LM Studio server rejects.

Confirmed directly against the server:

- `tool_choice: "required"` (string) → ✅ works, model emits the tool call
- `tool_choice: {"type":"function","function":{"name":...}}` (object) → ❌ `400 Invalid tool_choice type: 'object'`

So the server supports forced tool-calling via the string enum; it just rejects the object form — identical to the DeepSeek/MiniMax quirk.

## Suggested fix

The limitation is a property of the **server**, not individual models, so keying on model ID (as DeepSeek/MiniMax do) doesn't generalize to arbitrary local model names. Better to make local OpenAI-compatible providers default to suppressing the object form. Two options:

- **A — provider-aware default:** have `openai_compatible` (and `ollama`) resolve to `supports_tool_choice=False` by default, so the existing suppression path in `with_structured_output` kicks in. (`response_format: json_schema` also works on LM Studio and could be the preferred method.)
- **B — self-healing retry:** catch the specific `400 Invalid tool_choice type: 'object'` and retry with `tool_choice` suppressed (still structured) instead of dropping to free text.

Option A is the smaller change and mirrors the existing capability-table approach. This is complementary to #1051 (which hardens the fallback when a model skips the tool call) — A prevents the 400; #1051 handles the residual None case.

## Workaround

Adding the local model IDs to `_BY_ID` in `capabilities.py` with `supports_tool_choice=False` resolves it locally (verified end-to-end against LM Studio with `qwen3-coder-30b` — no more 400s, structured output succeeds), but per-ID entries don't scale to arbitrary local model names, hence the provider-level suggestion above.

## Environment

- Provider: `openai_compatible` → LM Studio server
- Model: `qwen3-coder-30b` (reproduces with any LM Studio chat model)
- macOS, Python 3.12

Ghraven:
## Summary
- add a `LocalCompatibleChatOpenAI` subclass for generic OpenAI-compatible local endpoints
- use it for the `openai_compatible` provider so structured-output binding suppresses LangChain's object-form `tool_choice`
- add a regression test that verifies local structured output still binds the schema as a tool

## Problem
LM Studio and similar local OpenAI-compatible servers can reject LangChain's function-spec object form of `tool_choice` even though they support function tools. Because arbitrary local model IDs fall through to the default capability row, structured-output calls can 400 and then fall back to free text.

Closes #1057.

## Before / After
Before: `openai_compatible` used `NormalizedChatOpenAI`, so unknown local model IDs kept the default `supports_tool_choice=True` behavior and could send object-form `tool_choice`.

After: `openai_compatible` uses `LocalCompatibleChatOpenAI`, which suppresses object-form `tool_choice` for function-calling structured output while still binding the schema as a tool.

## Verification
- `python -m compileall tradingagents/llm_clients/openai_client.py tests/test_openai_compatible_provider.py`
- `git diff --check`
- `python -m pytest tests/test_openai_compatible_provider.py tests/test_capabilities.py -q` (`28 passed`)

### DIFF
--- a/tests/test_openai_compatible_provider.py
+++ b/tests/test_openai_compatible_provider.py
@@ -6,6 +6,7 @@ model name is accepted, and the env backend URL precedence (#978).
 """
 
 import pytest
+from pydantic import BaseModel
 
 from tradingagents.llm_clients.api_key_env import get_api_key_env
 from tradingagents.llm_clients.factory import create_llm_client
@@ -36,13 +37,33 @@ def test_keyless_local_uses_placeholder_and_chat_completions(monkeypatch):
     llm = create_llm_client(
         provider="openai_compatible", model="qwen2.5", base_url="http://localhost:8000/v1"
     ).get_llm()
-    assert type(llm).__name__ == "NormalizedChatOpenAI"
+    assert type(llm).__name__ == "LocalCompatibleChatOpenAI"
     assert str(llm.openai_api_base) == "http://localhost:8000/v1"
     # keyless local servers: a placeholder key is sent
     key = llm.openai_api_key.get_secret_value() if hasattr(llm.openai_api_key, "get_secret_value") else llm.openai_api_key
     assert key == "EMPTY"
     # must use Chat Completions, not OpenAI's Responses API
     assert getattr(llm, "use_responses_api", False) in (False, None)
+
+
+@pytest.mark.unit
+def test_local_structured_output_suppresses_object_tool_choice(monkeypatch):
+    class Pick(BaseModel):
+        action: str
+
+    monkeypatch.delenv("OPENAI_COMPATIBLE_API_KEY", raising=False)
+    llm = create_llm_client(
+        provider="openai_compatible", model="qwen3-coder-30b", base_url="http://localhost:1234/v1"
+    ).get_llm()
+    bound = llm.with_structured_output(Pick)
+    first = bound.steps[0] if hasattr(bound, "steps") else bound
+    kwargs = getattr(first, "kwargs", {})
+
+    assert kwargs.get("tool_choice") is None or "tool_choice" not in kwargs
+    assert any(
+        tool.get("function", {}).get("name") == "Pick"
+        for tool in kwargs.get("tools", [])
+    ), f"schema not bound: {kwargs.get('tools', [])}"
 
 
 @pytest.mark.unit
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @Ghraven for the local-server tool_choice fix. Shipped in `517eeaf`: the generic openai_compatible provider now binds the schema as a tool without forcing object-form tool_choice, so structured output works on LM Studio and similar local servers.

PR #1062, opened 2026-06-20
fix(llm): suppress object-form tool_choice for local OpenAI-compatible

DogInfantry:
…e servers

Local OpenAI-compatible servers (LM Studio, llama.cpp, older vLLM, Ollama) reject the function-spec tool_choice object that langchain's function_calling structured-output path sends by default, returning HTTP 400. They run arbitrary model names the per-model capability table can't key off, so suppress tool_choice at the endpoint level for the openai_compatible and ollama providers. The schema is still bound as a tool, matching the existing DeepSeek/MiniMax handling.

Adds tests/test_local_server_tool_choice.py and a CHANGELOG entry.

Refs #1057


Yijia-Xiao:
Thanks @DogInfantry. Fixed in 8ac4371: the Ollama provider uses the local-compatible client, so structured output no longer sends the object-form tool_choice.

PR #1067, opened 2026-06-22
feat(dataflows): AKShare vendor for Chinese A-share stocks (.SS/.SZ)

ydhawesome:
## Summary

- **New vendor `akshare`** that routes `.SS` (Shanghai) and `.SZ` (Shenzhen) tickers to Chinese domestic data sources (东方财富 / 同花顺) instead of yfinance/Alpha Vantage.
- **Zero-config auto-routing**: `route_to_vendor()` detects A-share tickers by suffix and prepends `akshare` to the vendor chain automatically — no user configuration required; all other tickers are completely unaffected.
- **Sentiment analyst** branches on `is_a_share()` to substitute 东方财富 heat-rank for StockTwits and a clear placeholder for Reddit (not applicable for A-shares).
- **Optional install group**: `pip install tradingagents[china]` pulls in `akshare>=1.14.0` and `curl_cffi>=0.7.0`.
- **9 unit tests** covering `is_a_share()` detection and automatic vendor pre-routing.

## Data sources used (all public, no auth)

| Data type | Source |
|---|---|
| OHLCV price | `ak.stock_zh_a_hist` (东方财富, front-adjusted) |
| Technical indicators | stockstats (same pipeline as yfinance path) |
| News | 东方财富 JSONP search API (direct HTTP, avoids broken `ak.stock_news_em`) |
| Fundamentals / financials | `ak.stock_financial_abstract_ths` (同花顺) |
| Announcements | 东方财富 announcement REST API |
| Social sentiment | 东方财富 hot-rank (`ak.stock_hot_rank_em`) |

Note: `ak.stock_news_em` and several `stock_*_sheet_by_report_em` functions fail with PyArrow/pandas 2.x due to regex incompatibility and changed API response shapes. This PR works around both issues with direct HTTP calls to the underlying endpoints.

## Test plan

- [x] `pytest tests/test_akshare_routing.py -v` — 9/9 pass
- [x] End-to-end analysis of `600519.SS` (贵州茅台) completes successfully with MiniMax-M3, all data from domestic APIs, final decision returned in Chinese
- [x] `AAPL` analysis route is unchanged (verified no akshare code is invoked)

## Notes

- `curl_cffi` is used only for the 东方财富 news JSONP endpoint; the rest of the module uses `requests` (already in core deps).
- The `[china]` extra is intentionally separate so users in non-China markets don't pull in the additional deps.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,13 @@ dev = [
 bedrock = [
     "langchain-aws>=1.5.0",
 ]
+# Chinese A-share data (东方财富 / 同花顺). Optional because it pulls extra
+# data-scraping deps not needed for non-China use cases.
+# pip install "tradingagents[china]"
+china = [
+    "akshare>=1.14.0",
+    "curl_cffi>=0.7.0",
+]
 
 [project.scripts]
 tradingagents = "cli.main:app"
--- a/tests/test_akshare_routing.py
+++ b/tests/test_akshare_routing.py
@@ -2,6 +2,7 @@
 the akshare vendor first, while non-A-share tickers keep the configured path.
 """
 import copy
+import os
 import unittest
 from unittest import mock
 
@@ -10,7 +11,11 @@ import pytest
 import tradingagents.dataflows.config as config_module
 import tradingagents.default_config as default_config
 from tradingagents.dataflows import interface
-from tradingagents.dataflows.akshare_utils import is_a_share
+from tradingagents.dataflows.akshare_utils import (
+    _DOMESTIC_HOSTS,
+    _ensure_domestic_no_proxy,
+    is_a_share,
+)
 
 
 def _reset_config():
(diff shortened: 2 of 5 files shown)


ydhawesome:
Thanks for the review! Addressed all of the Gemini feedback and pushed two follow-up commits.

**1. Review feedback addressed**
- `get_stock_data_akshare`, `get_indicators_akshare`, and `_fetch_em_news` now re-raise on failure instead of returning a placeholder string, so `route_to_vendor`'s fallback chain works correctly.
- A-share tickers always move `akshare` to the front of the vendor list (even when a user config like `yfinance,akshare` already contains it).
- Removed the redundant `startswith(688)` check in `_em_prefix` and the no-op `interface.set_config` assignment in the test.

**2. Proxy-bypass fix for domestic data hosts**
- Found while testing in China: when a VPN/proxy (e.g. Clash) is active, the TLS stream to 东方财富/同花顺 gets corrupted (`SSL: DECRYPTION_FAILED_OR_BAD_RECORD_MAC`) or dropped by anti-crawler when reached from an overseas IP.
- Fix extends `NO_PROXY` with the domestic data hosts at import (covers akshare's internal bare `requests.get` too), plus explicit no-proxy overrides for the curl_cffi news call and the urllib announcements call. Documented limitation: this handles the common env-var proxy case but cannot defeat a system-wide TUN/transparent proxy (that must be fixed in the VPN config).

**Testing**
- `pytest tests/test_akshare_routing.py` — 12/12 pass (routing + NO_PROXY merge behavior).
- End-to-end: full analysis of `600519.SS` (贵州茅台) completes with all data from domestic APIs; non-A-share tickers (e.g. AAPL) never invoke akshare.

I will also rebase this branch onto the latest `main` shortly to resolve drift with the recent `interface.py` refactor. Happy to adjust anything — e.g. make the `NO_PROXY` mutation opt-in via config rather than applied at import.

gyx09212214-prog (commented):
Nice direction. The A-share routing is a real gap, and keeping the extra dependencies behind `[china]` is a good boundary.

One edge case I?d suggest checking: `get_global_news` appears to use `curr_date` as its first argument rather than a ticker. Since `route_to_vendor()` infers A-share status from `args[0]`, `get_global_news("2026-06-25", ...)` will not be detected as A-share and may never route to the new `akshare` implementation. If global market news is intended to use domestic sources for China workflows, this may need method-specific routing rather than ticker-based routing.

Also, `_ensure_domestic_no_proxy()` mutates `NO_PROXY` at import time. That may surprise users in non-China flows that only import the module. A lazy call inside the akshare vendor path might keep the side effect narrower.

PR #1070, opened 2026-06-22
refactor: remove setuptools from runtime deps and replace config deepcopy with read-only view

jaylew20250206:
## Summary

Two high-priority improvements identified in the project analysis report:

### 1. `pyproject.toml` — Move `setuptools` out of runtime dependencies

`setuptools>=80.9.0` was listed under `[project] dependencies` but it is a build-time tool, not a runtime library. It was already correctly declared in `[build-system] requires = ["setuptools>=61.0"]`, so this line was redundant.

**Impact**: `pip install tradingagents` no longer pulls in `setuptools` unnecessarily, reducing installation footprint and avoiding potential version conflicts.

### 2. `config.py` — Replace `deepcopy` with a deep-frozen read-only view

`get_config()` previously called `deepcopy(_config)` on every read — an O(n) copy of the entire configuration dictionary. This has been replaced with a recursive `MappingProxyType` freeze via the new `_deep_freeze()` helper.

**Benefits**:
- **Safety**: The returned view is deeply immutable — any attempt to mutate it (or any nested dict/list) raises `TypeError` immediately, catching bugs early instead of silently modifying a throwaway copy
- **Performance**: Avoids the overhead of `deepcopy` on every read (the freeze is still O(n) but avoids recursive object copying)
- **Type safety**: All functions now have full `dict[str, Any]` / `MappingProxyType[str, Any]` annotations

### Testing

All **491 tests pass**, 0 failures, 1 skipped (unrelated missing `langchain_aws` module).

## Changes

| File | Change |
|------|--------|
| `pyproject.toml` | Remove `setuptools>=80.9.0` from `[project] dependencies` |
| `tradingagents/dataflows/config.py` | Add `_deep_freeze()`, switch `get_config()` to `MappingProxyType`, add type annotations |
| `tests/test_dataflows_config.py` | Update test for read-only view semantics (verify `TypeError` on mutation) |

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,6 @@ dependencies = [
     "requests>=2.32.4",
     "rich>=14.0.0",
     "typer>=0.21.0",
-    "setuptools>=80.9.0",
     "stockstats>=0.6.5",
     "tqdm>=4.67.1",
     "typing-extensions>=4.14.0",
--- a/tests/test_dataflows_config.py
+++ b/tests/test_dataflows_config.py
@@ -2,6 +2,7 @@
 
 import copy
 import unittest
+from types import MappingProxyType
 
 import pytest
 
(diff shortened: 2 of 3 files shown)


Yijia-Xiao:
Thanks @jaylew20250206. The setuptools half is done in ed6eae4; the config copy semantics are deliberate, so that half is not taken.

PR #1074, opened 2026-06-22
feat(engine): retry an undecodable JSON response body instead of aborting the run

Lotus2077:
## What & why

A single transient response from a provider — a dropped connection, a timeout, a 5xx/429, or a **malformed / non-JSON body** — currently propagates out of a graph node and aborts the entire run. On a long multi-agent run that's expensive, and `checkpoint_enabled` only bounds the loss (resume from the last step) — it doesn't prevent the crash.

This wraps `NormalizedChatOpenAI.invoke` in a small retry loop over a curated transient-error set, so one bad response is retried instead of fatal.

## What it does

- Retries a curated transient set: `APITimeoutError`, `APIConnectionError`, `InternalServerError`, `RateLimitError`, `APIResponseValidationError`, and `json.JSONDecodeError`.
  - The OpenAI SDK already retries the connection/timeout/5xx/429 family *inside a single call*, but **not** response-parse failures — the malformed/non-JSON body is the main case this adds at the invoke level.
- **Permanent 4xx** errors (auth, bad request, not found) are deliberately excluded — they still fail fast.
- Bounded by the existing `max_retries` field (default 2) with capped exponential backoff; each retry logs to stderr.
- Two optional config knobs, forwarded via `_get_provider_kwargs`:
  - `llm_max_retries` / `TRADINGAGENTS_LLM_MAX_RETRIES`
  - `llm_timeout` / `TRADINGAGENTS_LLM_TIMEOUT`
  - Unset → SDK defaults (behavior unchanged).

## Tests

`tests/test_llm_retry.py` covers: a transient parse failure retries and recovers; a permanent error surfaces immediately without retry; a persistent transient exhausts `max_retries` before re-raising.

## Scope

Engine-only, opt-in, backward-compatible. 6 files, +155/-1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 Breaking changes within the 0.x line are called out explicitly.
 
+## [Unreleased]
+
+### Added
+
+- **Engine: a 2xx response with an undecodable JSON body is retried instead of
+  aborting the run.** The OpenAI SDK already retries the connection / timeout /
+  5xx / 429 family internally, but it parses each response body only once —
+  *after* that retry loop — so a successful 2xx whose `application/json` body is
+  truncated or garbled (`json.JSONDecodeError`) would propagate out of a node and
+  crash a long graph run (checkpointing bounds the loss, it doesn't prevent the
+  crash). `NormalizedChatOpenAI.invoke` now retries that one case, bounded by
+  `max_retries` (default 2) with capped exponential backoff; the HTTP-transient
+  family is left to the SDK (so retries aren't multiplied) and permanent 4xx
+  errors still fail fast. New config knobs `llm_max_retries` / `llm_timeout` (env
+  `TRADINGAGENTS_LLM_MAX_RETRIES` / `TRADINGAGENTS_LLM_TIMEOUT`). (Malformed LLM
+  *structured-output* content is a separate layer, recovered by the per-agent
+  free-text fallback.)
+
 ## [0.3.0] — 2026-06-22
 
 Stabilization and extensibility release: a CI gate, a unified verified
--- a/README.md
+++ b/README.md
@@ -268,6 +268,21 @@ ta = TradingAgentsGraph(config=config)
 _, decision = ta.propagate("NVDA", "2026-01-15")
 ```
 
+### Transient-failure retries
+
+The OpenAI SDK already retries dropped connections, timeouts, and 5xx/429 — but it
+parses each response body only once, *after* that retry loop, so a successful 2xx
+whose `application/json` body is truncated or garbled (`json.JSONDecodeError`) would
+otherwise abort a whole run. `NormalizedChatOpenAI.invoke` retries that one case,
+bounded by `max_retries` with capped exponential backoff; the HTTP-transient family
+is left to the SDK (no multiplied retries) and permanent errors (auth, bad request)
+still fail fast. Tune it with `llm_max_retries` / `llm_timeout` in your config, or
+`TRADINGAGENTS_LLM_MAX_RETRIES` / `TRADINGAGENTS_LLM_TIMEOUT` in `.env` (defaults: 2
+retries, no explicit timeout). (A model emitting malformed structured-output content
+is handled separately, by the per-agent free-text fallback.) This complements
+`--checkpoint`: checkpointing limits how much you lose on a crash; this prevents the
+crash in the first place.
+
 ## Reproducibility
 
 TradingAgents is LLM-driven, so two runs of the same ticker and date can differ. This is expected for a research tool built on language models, not a defect. The variation comes from a few distinct sources, and it helps to separate them.
--- a/tests/test_llm_retry.py
+++ b/tests/test_llm_retry.py
@@ -1,0 +1,89 @@
+"""Unit tests for ``NormalizedChatOpenAI.invoke`` transient-failure retry.
+
+The retry loop wraps ``super().invoke()`` and retries only an undecodable JSON
+response body (``json.JSONDecodeError``) — the one transient failure the OpenAI
+SDK does not retry itself. The HTTP-transient family is delegated to the SDK,
+and permanent errors surface immediately. These tests exercise that loop
+directly by stubbing the parent ``invoke``.
+"""
+
+import json
+
+import httpx
+import openai
+import pytest
+
+import tradingagents.llm_clients.openai_client as oc
+
+
+@pytest.fixture(autouse=True)
+def _fast_and_isolated(monkeypatch):
+    # Don't actually sleep between retries, and don't require a real message
+    # object back from the (stubbed) parent invoke.
+    monkeypatch.setattr(oc.time, "sleep", lambda *_a, **_k: None)
+    monkeypatch.setattr(oc, "normalize_content", lambda value: value)
+
+
+def _client(max_retries):
+    # Build an instance without network/credentials; we only exercise the retry
+    # loop, which reads ``self.max_retries`` and calls ``super().invoke()``.
+    return oc.NormalizedChatOpenAI.model_construct(max_retries=max_retries)
+
+
+def test_transient_failure_is_retried_then_succeeds(monkeypatch):
+    calls = []
+
+    def flaky(self, *args, **kwargs):
+        calls.append(1)
+        if len(calls) == 1:
+            raise json.JSONDecodeError("Expecting value", "", 0)
+        return "recovered"
+
+    monkeypatch.setattr(oc.ChatOpenAI, "invoke", flaky)
+    assert _client(max_retries=2).invoke("prompt") == "recovered"
+    assert len(calls) == 2  # one transient failure, then success
+
+
+def test_permanent_error_is_not_retried(monkeypatch):
+    calls = []
+
+    def boom(self, *args, **kwargs):
+        calls.append(1)
+        raise ValueError("permanent 4xx-style failure")
+
+    monkeypatch.setattr(oc.ChatOpenAI, "invoke", boom)
+    with pytest.raises(ValueError):
+        _client(max_retries=5).invoke("prompt")
+    assert len(calls) == 1  # surfaced immediately, never retried
+
+
+def test_persistent_transient_exhausts_budget_then_raises(monkeypatch):
+    calls = []
+
+    def always_transient(self, *args, **kwargs):
+        calls.append(1)
+        raise json.JSONDecodeError("Expecting value", "", 0)
+
+    monkeypatch.setattr(oc.ChatOpenAI, "invoke", always_transient)
+    with pytest.raises(json.JSONDecodeError):
+        _client(max_retries=2).invoke("prompt")
+    assert len(calls) == 3  # initial attempt + max_retries
+
+
+def test_http_transient_is_delegated_to_the_sdk_not_outer_retried(monkeypatch):
+    # The OpenAI SDK retries the connection / timeout / 5xx / 429 family itself
+    # (max_retries is forwarded to it), so the invoke loop must NOT also catch
+    # those — that would multiply requests. An HTTP-family error therefore
+    # propagates on the first attempt instead of being re-retried here.
+    calls = []
+
+    def http_transient(self, *args, **kwargs):
+        calls.append(1)
+        raise openai.APITimeoutError(
+            request=httpx.Request("POST", "https://api.openai.com/v1/chat/completions")
+        )
+
+    monkeypatch.setattr(oc.ChatOpenAI, "invoke", http_transient)
+    with pytest.raises(openai.APITimeoutError):
+        _client(max_retries=3).invoke("prompt")
+    assert len(calls) == 1  # delegated to the SDK, not re-retried at invoke level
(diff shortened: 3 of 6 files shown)


Lotus2077:
Thanks for the careful review — all three were on the mark, and I've pushed an update.

**1. Multiplicative retries — fixed.** Confirmed: `max_retries` is forwarded to the OpenAI SDK client, and the SDK already retries the connection/timeout/5xx/429 family internally, so catching those same classes in the outer loop squared the budget (~9 requests at the default). I narrowed the outer loop to only what the SDK does *not* retry and left the HTTP-transient family entirely to the SDK (keeping its Retry-After/backoff/jitter). Added a test asserting an HTTP-family error isn't re-retried at the invoke level.

**2. Structured-output parsing — you're right; it's handled a layer up.** `with_structured_output` returns `llm | output_parser`, and the parser runs outside `NormalizedChatOpenAI.invoke`, so the loop legitimately doesn't wrap it. Those content-parse failures are already caught by the agents' `invoke_structured_or_freetext` free-text fallback at the four structured sites; the non-structured researcher/risk nodes have no fallback, which is exactly where this invoke-level retry is the only net. I've scoped the retry — and the docstring/CHANGELOG/README — to the *transport* case only: a 2xx whose `application/json` body fails to decode (`json.JSONDecodeError`), which the SDK surfaces inside `invoke` but doesn't retry.

**3. logging instead of print — done.** Switched to `logging.getLogger(__name__).warning(...)`, matching `agents/utils/structured.py`.

While verifying, I also dropped `openai.APIResponseValidationError` from the set: the SDK only raises it under `_strict_response_validation=True`, which this codebase doesn't enable, so it was unreachable here. The retry set is now just `(json.JSONDecodeError,)`.

praxstack:
_Automated principal-engineer review (findings adversarially verified before posting)._

Thanks for tackling this — wrapping `NormalizedChatOpenAI.invoke` for undecodable-JSON 2xx bodies closes a real gap the OpenAI SDK does not cover (`response.json()` raises a raw `json.JSONDecodeError` after the SDK's own retry loop). The premise, scope, and most of the implementation are sound. A couple of things, though, mean the feature is a silent no-op in the default configuration.

## High — correctness

### Retry is a no-op under the default config
**File:** `tradingagents/llm_clients/openai_client.py`

The invoke-level retry budget is derived from `self.max_retries`, but on a `NormalizedChatOpenAI` built the way production builds it, that langchain field defaults to `None` (the `2` lives only on the underlying SDK HTTP client `root_client.max_retries` and is never written back to the langchain field). `DEFAULT_CONFIG['llm_max_retries']` is also `None`, and `_get_provider_kwargs` only forwards `max_retries` when it's not `None`/`''`. Net effect: out of the box, `attempts == 1`, so the new retry loop never retries anything.

**Failure scenario:** A user running the shipped defaults hits a transient undecodable-JSON 2xx body — the exact case this PR targets — and gets the same hard failure as before, because the loop executes exactly one attempt.

**Suggested fix:** Decouple the invoke-level retry budget from the langchain `max_retries` field (or default it explicitly), e.g.
```python
attempts = max(self.max_retries if self.max_retries is not None else 2, 0) + 1
```
or read a dedicated retry-count setting. Then add a test (see below) that exercises the unset-default path.

## Medium — tests

### Tests construct the client with an explicit `max_retries`, masking the default-config bug
**File:** `tests/test_llm_retry.py`

All four tests use `_client(max_retries=N)` via `model_construct(max_retries=...)`, which forces a non-`None` value and bypasses pydantic field defaults. None of them exercises the real default path (`max_retries` unset → `None`) — which is precisely where the feature does nothing. The green suite proves the loop works *when a budget is set*, not that the shipped default retries anything; `model_construct` also doesn't model the production construction path (`chat_cls(**llm_kwargs)` with full validation).

**Failure scenario:** The default-config no-op above ships undetected because every test pre-seeds a budget the production default never sets.

**Suggested fix:** Add a test that builds the client the way production does (no explicit `max_retries`, full validation) and asserts the JSON-decode case is actually retried under the documented default — assert the concrete attempt count.

## Low — docs

### Docs understate scope: structured-output / tool-calling chains *are* routed through the retry
**File:** `tradingagents/llm_clients/openai_client.py`

The comment and README state that malformed structured-output is "a separate layer … outside invoke." In practice `with_structured_output` returns a `RunnableSequence` and `bind_tools` returns a `RunnableBinding`, both of which delegate back into `NormalizedChatOpenAI.invoke` (`RunnableBinding.invoke` → `self.bound.invoke`; `RunnableSequence.invoke` → per-step `step.invoke`). So a body-level `json.JSONDecodeError` on a structured/tool call *is* caught and retried here. Harmless (broader coverage than claimed), but the documented mental model is inaccurate and could mislead future maintainers.

**Suggested fix:** Reword to note that invoke-level retries cover bare `invoke`, tool-calling (`bind_tools`), and structured-output chains alike, since langchain's `RunnableBinding`/`RunnableSequence` delegate to `invoke`. Clarify that the free-text fallback is for malformed *content* that decodes fine as JSON but fails schema validation downstream — not for body-level `JSONDecodeError`.

## Overall

Good direction and a genuinely useful gap to close. The blocker is the default-config no-op: with the budget defaulting to `None`, the feature ships inert and the test suite gives false confidence. Fix the budget derivation, add a default-path test, and tighten the docs, and this is ready to land.

🤖 Automated review via Claude Code

praxstack:
_Proposed implementation of the fixes from my earlier automated review, offered as a ready-to-apply patch. Generated by an automated principal-engineer pass; please treat as a starting point._

**What it does:** Fix the LLM retry-budget no-op under default config; +3 tests

The diff is **against this PR's head commit** (apply on top of the PR branch, not `main`):

```bash
git apply pr-1074.patch   # from the repo root, on this PR's branch
```

<details><summary>📋 pr-1074.patch (click to expand)</summary>

```diff
diff --git c/README.md w/README.md
index 62501d9..29d50b9 100644
--- c/README.md
+++ w/README.md
@@ -273,13 +273,17 @@ _, decision = ta.propagate("NVDA", "2026-01-15")
 The OpenAI SDK already retries dropped connections, timeouts, and 5xx/429 — but it
 parses each response body only once, *after* that retry loop, so a successful 2xx
 whose `application/json` body is truncated or garbled (`json.JSONDecodeError`) would
-otherwise abort a whole run. `NormalizedChatOpenAI.invoke` retries that one case,
-bounded by `max_retries` with capped exponential backoff; the HTTP-transient family
-is left to the SDK (no multiplied retries) and permanent errors (auth, bad request)
-still fail fast. Tune it with `llm_max_retries` / `llm_timeout` in your config, or
-`TRADINGAGENTS_LLM_MAX_RETRIES` / `TRADINGAGENTS_LLM_TIMEOUT` in `.env` (defaults: 2
-retries, no explicit timeout). (A model emitting malformed structured-output content
-is handled separately, by the per-agent free-text fallback.) This complements
+otherwise abort a whole run. `NormalizedChatOpenAI.invoke` retries that one case with
+capped exponential backoff; the HTTP-transient family is left to the SDK (no
+multiplied retries) and permanent errors (auth, bad request) still fail fast. Because
+langchain's tool-calling (`bind_tools`) and structured-output (`with_structured_output`)
+wrappers both delegate back into `invoke`, this retry covers bare, tool-calling, and
+structured-output calls alike. Tune the budget with `llm_max_retries` / `llm_timeout`
+in your config, or `TRADINGAGENTS_LLM_MAX_RETRIES` / `TRADINGAGENTS_LLM_TIMEOUT` in
+`.env` (defaults: 2 retries, no explicit timeout). (A different case — content that
+decodes fine as JSON but fails downstream schema validation, not a body-level
+`JSONDecodeError` — is handled separately, by the per-agent free-text fallback.)
+This complements
 `--checkpoint`: checkpointing limits how much you lose on a crash; this prevents the
 crash in the first place.
 
diff --git c/tests/test_llm_retry.py w/tests/test_llm_retry.py
index ccf396a..d79d0a7 100644
--- c/tests/test_llm_retry.py
+++ w/tests/test_llm_retry.py
@@ -87,3 +87,54 @@ def test_http_transient_is_delegated_to_the_sdk_not_outer_retried(monkeypatch):
     with pytest.raises(openai.APITimeoutError):
         _client(max_retries=3).invoke("prompt")
     assert len(calls) == 1  # delegated to the SDK, not re-retried at invoke level
+
+
+def test_default_config_path_still_retries(monkeypatch):
+    # Production builds the client without forwarding ``max_retries``: the shipped
+    # ``DEFAULT_CONFIG['llm_max_retries']`` is ``None``, so ``_get_provider_kwargs``
+    # omits the kwarg and the field is left unset (``None``). The invoke-level retry
+    # must NOT be derived from that field (it would be a single-attempt no-op);
+    # it falls back to ``_DEFAULT_INVOKE_RETRIES`` so the feature works out of the box.
+    calls = []
+
+    def flaky(self, *args, **kwargs):
+        calls.append(1)
+        if len(calls) == 1:
+            raise json.JSONDecodeError("Expecting value", "", 0)
+        return "recovered"
+
+    monkeypatch.setattr(oc.ChatOpenAI, "invoke", flaky)
+    # max_retries unset (None) is exactly what the documented default leaves it as.
+    assert _client(max_retries=None).invoke("prompt") == "recovered"
+    assert len(calls) == 2  # one transient failure was retried under the default budget
+
+
+def test_default_budget_exhausts_after_default_retries(monkeypatch):
+    # Under the unset default, a persistently undecodable body is retried exactly
+    # ``_DEFAULT_INVOKE_RETRIES`` times before surfacing (initial attempt + retries).
+    calls = []
+
+    def always_transient(self, *args, **kwargs):
+        calls.append(1)
+        raise json.JSONDecodeError("Expecting value", "", 0)
+
+    monkeypatch.setattr(oc.ChatOpenAI, "invoke", always_transient)
+    with pytest.raises(json.JSONDecodeError):
+        _client(max_retries=None).invoke("prompt")
+    assert len(calls) == oc._DEFAULT_INVOKE_RETRIES + 1
+
+
+def test_explicit_zero_retries_disables_invoke_retry(monkeypatch):
+    # An explicit ``llm_max_retries=0`` must be honored as "no invoke-level retry"
+    # (one attempt only) — it must not be confused with the unset default that
+    # falls back to ``_DEFAULT_INVOKE_RETRIES``.
+    calls = []
+
+    def always_transient(self, *args, **kwargs):
+        calls.append(1)
+        raise json.JSONDecodeError("Expecting value", "", 0)
+
+    monkeypatch.setattr(oc.ChatOpenAI, "invoke", always_transient)
+    with pytest.raises(json.JSONDecodeError):
+        _client(max_retries=0).invoke("prompt")
+    assert len(calls) == 1  # explicit 0 means a single attempt, no retry
diff --git c/tradingagents/llm_clients/openai_client.py w/tradingagents/llm_clients/openai_client.py
index 4199645..0ec37d1 100644
--- c/tradingagents/llm_clients/openai_client.py
+++ w/tradingagents/llm_clients/openai_client.py
@@ -27,11 +27,29 @@ logger = logging.getLogger(__name__)
 # the SDK so retries aren't multiplied, and permanent 4xx errors (auth, bad
 # request) still fail fast.
 #
-# Not handled here (different layer): a model emitting malformed *structured-
-# output* content surfaces in the downstream parser, outside invoke, and is
-# recovered by the agents' free-text fallback (agents/utils/structured.py).
+# This retry is reached for a bare ``invoke`` and equally for tool-calling
+# (``bind_tools`` → ``RunnableBinding.invoke`` → ``self.bound.invoke``) and
+# structured-output (``with_structured_output`` → ``RunnableSequence.invoke`` →
+# per-step ``step.invoke``) chains, since both delegate back into this method:
+# a body-level ``JSONDecodeError`` on any of them is caught and retried here.
+#
+# Separate layer (NOT this retry): a model that emits *content* which decodes
+# fine as JSON but fails downstream schema validation surfaces in the parser
+# outside invoke and is recovered by the agents' free-text fallback
+# (agents/utils/structured.py) — that is malformed content, not a body-level
+# JSONDecodeError.
 _TRANSIENT_LLM_ERRORS: tuple[type[Exception], ...] = (json.JSONDecodeError,)
 
+# Invoke-level retry budget when no ``max_retries`` is configured. The langchain
+# ``ChatOpenAI.max_retries`` field is NOT a reliable source for this: under the
+# shipped defaults (``DEFAULT_CONFIG['llm_max_retries']`` is ``None``, so
+# ``_get_provider_kwargs`` never forwards it) the field is left at langchain's
+# own default, and the SDK's HTTP retry count lives on ``root_client.max_retries``
+# — neither is guaranteed to be a usable budget for this loop. So derive the
+# budget explicitly and fall back to this default, ensuring the retry is active
+# out of the box rather than a silent no-op.
+_DEFAULT_INVOKE_RETRIES = 2
+
 
 class NormalizedChatOpenAI(ChatOpenAI):
     """ChatOpenAI with normalized content output and capability-aware binding.
@@ -54,10 +72,13 @@ class NormalizedChatOpenAI(ChatOpenAI):
 
     def invoke(self, input, config=None, **kwargs):
         # Retry an undecodable-JSON response body instead of letting it abort the
-        # graph (see _TRANSIENT_LLM_ERRORS). `max_retries` (the ChatOpenAI field,
-        # default 2) bounds the attempts and `timeout` bounds each request; both
-        # come from the `llm_max_retries` / `llm_timeout` config keys (TRADINGAGENTS_LLM_*).
-        attempts = max(self.max_retries or 0, 0) + 1
+        # graph (see _TRANSIENT_LLM_ERRORS). The budget comes from the configured
+        # `llm_max_retries` (TRADINGAGENTS_LLM_MAX_RETRIES) when set, else
+        # `_DEFAULT_INVOKE_RETRIES` — NOT `self.max_retries`, which is `None` under
+        # the shipped defaults and would make this loop a single-attempt no-op.
+        # `timeout` (the `llm_timeout` config key) bounds each request.
+        retries = self.max_retries if self.max_retries is not None else _DEFAULT_INVOKE_RETRIES
+        attempts = max(retries, 0) + 1
         for attempt in range(attempts):
             try:
                 return normalize_content(super().invoke(input, config, **kwargs))
```

</details>

🤖 Automated patch via Claude Code — not pushed anywhere; yours to apply, adapt, or ignore.

PR #1075, opened 2026-06-22
Normalize benchmark ticker before fetching return history

gyx09212214-prog:
## Summary
- Normalize configured benchmark tickers before fetching return history in `_fetch_returns`
- Add a regression test covering broker/CFD benchmark aliases such as `SPX500 -> ^GSPC`

## Why
`_fetch_returns` already normalizes the analyzed ticker before querying yfinance, but it passed the configured benchmark directly to `yf.Ticker(...)`. When users set `benchmark_ticker` to an alias such as `SPX500`, yfinance receives the raw alias instead of the canonical Yahoo symbol `^GSPC`, so realized alpha can fail or be computed against missing benchmark data.

This keeps canonical benchmarks like `SPY` unchanged while allowing user-configured aliases to follow the same normalization path as traded symbols.

## Tests
- `python -m pytest tests/test_symbol_normalization_paths.py tests/test_symbol_utils.py tests/test_reporting.py -q`
- `python -m ruff check tradingagents/graph/trading_graph.py tests/test_symbol_normalization_paths.py`
- `git diff --check`

### DIFF
--- a/tests/test_symbol_normalization_paths.py
+++ b/tests/test_symbol_normalization_paths.py
@@ -55,6 +55,27 @@ def test_fetch_returns_normalizes_symbol(monkeypatch):
     assert raw is not None and days is not None
 
 
+def test_fetch_returns_normalizes_benchmark(monkeypatch):
+    queried = []
+
+    class FakeTicker:
+        def __init__(self, symbol):
+            queried.append(symbol)
+
+        def history(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @gyx09212214-prog. Fixed in 60dcf64: an explicitly configured benchmark is resolved through the same symbol mapping as the analyzed ticker.

PR #1081, opened 2026-06-23
fix bug:log file can't store non ascii encoding, eg chinese character

ZHUYAWEI:
full_states_log file can only store ascii character ,so non ascii character can't be interpret ， 
for example for chinese character “平均”,it stores as \u5E73\u5747


Yijia-Xiao:
Thanks @ZHUYAWEI. Fixed in 6398951: the state log is written without escaping non-ASCII, so reports stay readable.

PR #1082, opened 2026-06-24
feat(trader): add probability + risk/reward review to every proposal

ydhawesome:
## Summary

Every Trader proposal now ends with a **quantified, two-sided verdict** instead of a one-sided narrative:

- **Bull case + Bear case** (both required): the strongest arguments for and against, each grounded in specific evidence. The prompt explicitly tells the model to argue both sides and not to strawman the opposing view.
- **Win probability** (required, 0-100): the model commits to a likelihood estimate, with an instruction not to default to 50.
- **Deterministic risk/reward review**: the win probability comes from the model, but the **risk/reward ratio**, **expected value** (in R-multiples), and **breakeven win-rate** are computed in code from the entry/stop/target levels — so the numbers are always arithmetically consistent rather than hallucinated by the LLM.

### Example output

```
**Action**: Buy
**Reasoning**: ...
**Bull Case**: Datacenter demand accelerating; ...
**Bear Case**: Export-control overhang; ...
**Entry Price**: 189.5
**Stop Loss**: 178.0
**Target Price**: 212.0

### Probability & Risk/Reward
- **Win Probability**: 64%
- **Risk/Reward Ratio**: 2.00 : 1 (potential +22.50 vs risk -11.50)
- **Expected Value**: +0.92R (favorable)
- **Breakeven Win-Rate**: 33% (current 64% is above breakeven)

FINAL TRANSACTION PROPOSAL: **BUY**
```

When price levels are absent (e.g. a Hold), the risk/reward line degrades to `n/a` and the probability + bull/bear still render.

## Changes

- `tradingagents/agents/schemas.py`: extend `TraderProposal` with `bull_case` / `bear_case` / `win_probability` (required) and `target_price` (optional, covered by the existing nullish-float coercion validator); add `_render_trade_review()` and wire it into `render_trader_proposal()`.
- `tradingagents/agents/trader/trader.py`: prompt now instructs the model to argue both sides, commit a win probability, and supply entry/stop/target.
- `tests/test_structured_agents.py`: updated constructors for the new required fields; added tests for exact R/R, EV and breakeven math, favorable-vs-unfavorable EV flagging, the `n/a` path, and `target_price` nullish coercion.

## Test plan

- [x] `pytest tests/test_structured_agents.py` — 27 passed
- [x] `ruff check` on the changed files — clean
- [x] Structured-output JSON schema marks `bull_case` / `bear_case` / `win_probability` as required, so providers using json_schema mode (OpenAI/xAI/MiniMax) enforce them; the free-text fallback path is unchanged.

## Notes

The win-probability/R-R framing is deliberately model-agnostic and adds no new dependencies. Happy to adjust labels or make the review section opt-in via config if preferred.

### DIFF
--- a/tests/test_structured_agents.py
+++ b/tests/test_structured_agents.py
@@ -36,10 +36,20 @@ from tradingagents.agents.trader.trader import create_trader
 @pytest.mark.unit
 class TestRenderTraderProposal:
     def test_minimal_required_fields(self):
-        p = TraderProposal(action=TraderAction.HOLD, reasoning="Balanced setup; no edge.")
+        p = TraderProposal(
+            action=TraderAction.HOLD,
+            reasoning="Balanced setup; no edge.",
+            bull_case="Margins resilient.",
+            bear_case="Cash flow deteriorating.",
+            win_probability=50,
+        )
         md = render_trader_proposal(p)
         assert "**Action**: Hold" in md
         assert "**Reasoning**: Balanced setup; no edge." in md
+        # Bull/bear cases and the probability review are always present.
+        assert "**Bull Case**: Margins resilient." in md
+        assert "**Bear Case**: Cash flow deteriorating." in md
+        assert "**Win Probability**: 50%" in md
         # The trailing FINAL TRANSACTION PROPOSAL line is preserved for the
         # analyst stop-signal text and any external code that greps for it.
         assert "FINAL TRANSACTION PROPOSAL: **HOLD**" in md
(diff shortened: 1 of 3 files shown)


gyx09212214-prog (commented):
The deterministic risk/reward math should account for direction. Right now reward/risk uses `abs(target - entry)` and `abs(entry - stop)`, so a BUY with `target_price < entry_price` or `stop_loss > entry_price`, or a SELL with `target_price > entry_price` / `stop_loss < entry_price`, still renders a positive R/R and EV. That can make inverted levels look favorable.

Could we validate direction-specific ordering before computing R/R, or render n/a with a warning, and add tests for invalid BUY and SELL level ordering?

PR #1084, opened 2026-06-24
Add batch analysis CLI command

crabbag3:


Yijia-Xiao:
Thanks @crabbag3. Superseded by 8721b92: tradingagents backtest runs a grid of tickers and dates.

PR #1086, opened 2026-06-25
fix(smoke): clarify structured output markdown checks

### LINKED ISSUE #997: there are a few significant logical flaws and inconsistencies, primarily centering around how the script handles the data types it claims to process.
​1. Type Mismatch and Likely TypeError Crash
• ​The Issue: The script's docstring explicitly states that the agents return "the typed Pydantic instance + the rendered markdown for each". However, near the bottom of the script, the checks loop attempts to find string markdown headers (such as "**Recommendation**:" or "**Action**:") directly inside the returned variables using the syntax ok = marker in text.
• ​Why it fails: If investment_plan, trader_plan, and final_decision are actually Pydantic model instances (as the docstring suggests), evaluating marker in text will raise a Python TypeError. You cannot use the in operator to search for a substring directly inside a Pydantic object; it requires a string or an iterable.
• ​Contradictory Type Hints: Compounding this issue, the helper function _make_pm_state uses string type hints for these parameters (investment_plan: str, trader_plan: str), which actively contradicts the docstring's claim that they are Pydantic objects.
​2. Missing the Promised Pydantic Output
• ​The Issue: The script claims in its introductory documentation that it "prints the typed Pydantic instance + the rendered markdown for each".
• ​Why it fails: The code simply extracts a single key from each agent's result dictionary (e.g., investment_plan = rm_result["investment_plan"]) and passes it directly to _print_section. It never attempts to separate, extract, or independently print a Pydantic instance alongside a rendered markdown string. If the returned key merely holds a markdown string, the script fails its stated purpose of verifying the Pydantic instance.
​3. Unused Standard Library Import
• ​The Issue: The statement import os is declared at the top of the file but the os module is never actually utilized anywhere in the code. While this won't cause the script to crash at runtime, it is poor practice and will be flagged as an error by any standard Python linter (like Flake8 or Ruff).

gyx09212214-prog:
﻿## Summary

Clarify the structured-output smoke script so it describes and checks the actual agent API output: rendered markdown produced through the normal agent callables.

## Why

Issue #997 points out that `scripts/smoke_structured_output.py` says it prints typed Pydantic instances plus rendered markdown, while the graph-facing agent API returns rendered markdown strings. The script also checked markers directly with `marker in text`, which would raise a raw `TypeError` if a future change returned a non-string object.

## What changed

- Update the smoke script docstring to describe schema-backed rendered markdown instead of promising typed Pydantic instances.
- Add small marker-check helpers that coerce smoke outputs to text before checking required headers.
- Keep the existing structure checks and exit-code behavior unchanged.
- Add regression coverage for non-string structured-like output and PASS/FAIL accounting.

## Duplicate check

I checked current open PRs and searched for `smoke_structured_output`, `_make_pm_state`, and structured-output smoke fixes; I found no overlapping open PR.

## Validation

- `python -m pytest tests/test_smoke_structured_output.py tests/test_structured_agents.py -q`
- `python -m pytest tests/test_smoke_structured_output.py -q`
- `python -m py_compile scripts/smoke_structured_output.py tests/test_smoke_structured_output.py`
- `python -m ruff check scripts/smoke_structured_output.py tests/test_smoke_structured_output.py`
- `python -m ruff format --check scripts/smoke_structured_output.py tests/test_smoke_structured_output.py`
- `git diff --check`

Fixes #997.

### DIFF
--- a/scripts/smoke_structured_output.py
+++ b/scripts/smoke_structured_output.py
@@ -105,6 +105,8 @@ def _print_section(title: str, content: object) -> None:
 
 def _as_text(content: object) -> str:
     """Coerce smoke outputs to text before marker checks."""
+    if content is None:
+        return ""
     return content if isinstance(content, str) else str(content)
 
 
@@ -116,9 +118,9 @@ def _missing_markers(content: object, required: list[str]) -> list[str]:
 def _run_structure_checks(checks: list[tuple[str, object, list[str]]]) -> int:
     failures = 0
     for name, content, required in checks:
-        missing = set(_missing_markers(content, required))
+        text = _as_text(content)
         for marker in required:
-            ok = marker not in missing
+            ok = marker in text
             print(f"  {'PASS' if ok else 'FAIL'}  {name}: contains {marker!r}")
             failures += int(not ok)
     return failures
--- a/scripts/smoke_structured_output.py
+++ b/scripts/smoke_structured_output.py
@@ -1,11 +1,12 @@
 """End-to-end smoke for structured-output agents against a real LLM provider.
 
 Runs the three decision-making agents (Research Manager, Trader, Portfolio
-Manager) directly with their structured-output bindings and prints the
-typed Pydantic instance + the rendered markdown for each.  Use this to
-verify a provider's native structured-output mode (json_schema for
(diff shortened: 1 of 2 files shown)


Tauric-Research:

PR #1103, opened 2026-06-30
feat(bedrock): support native Bedrock API key (AWS_BEARER_TOKEN_BEDROCK)

praxstack:
## Summary

Adds first-class support for Amazon Bedrock's native **API keys** (bearer tokens, env `AWS_BEARER_TOKEN_BEDROCK`) as an alternative to the existing AWS SigV4 credential-chain auth — no AWS access keys, profile, or IAM role required.

The interesting finding while building this: **the bearer token already works at runtime with the currently-pinned dependency floor**, with zero client-code change. `langchain-aws` (≥1.5.0) reads `AWS_BEARER_TOKEN_BEDROCK` into its `bedrock_api_key` field and botocore (≥1.39.0) sends `Authorization: Bearer <token>` instead of SigV4. The existing `BedrockClient.get_llm()` already constructs `ChatBedrockConverse(model=…, region_name=…)` with no creds — exactly the supported bearer-only pattern.

So the real gap was **discoverability, UX, and tests**, not auth plumbing. This PR closes that gap without breaking the credential-chain path.

### Verified live (not just by reading source)
With only `AWS_BEARER_TOKEN_BEDROCK` set and all AWS access-key vars scrubbed, the existing client produces:
```
Authorization scheme: Bearer
Uses our bearer token: True
Is SigV4: False
```
against `langchain-aws==1.6.1` / `botocore==1.43.37`.

## Changes
- **`cli/utils.py`** — `ensure_api_key()` special-cases `bedrock` to **advise, never force-prompt**. It returns the bearer token if set, stays silent if the SigV4 chain looks configured (env-only heuristic, no boto3 import / IMDS call), and prints an actionable hint only when *neither* mode is detected — turning the otherwise-opaque deferred boto3 error into guidance. IAM-role / profile users are never nagged for a token they don't use.
- **`api_key_env.py`** — keeps `bedrock → None` (neither mode is a single forced key) and exports `BEDROCK_BEARER_TOKEN_ENV` as the canonical name, with rationale.
- **`bedrock_client.py`** — docstring now documents both auth modes and their precedence.
- **Docs** — `README.md`, `.env.example`, and the `[bedrock]` extra in `pyproject.toml` describe both modes and the version-floor rationale; `CHANGELOG.md` gets an `[Unreleased]` entry.

## Tests
- `test_construction_with_bearer_token_only` — bearer-only env constructs without raising **and** the token reaches botocore's request signer (asserts bearer auth, not a silent SigV4 fallback). Gated by `importorskip("langchain_aws")`.
- Three `ensure_api_key("bedrock")` tests cover all auth states (bearer set / chain configured / neither), each asserting the key prompt is **never** shown.
- Hardened the pre-existing `test_construction_when_extra_installed` to scrub ambient AWS env (it silently relied on a clean environment; installing the `[bedrock]` extra locally surfaced the fragility).

Full suite: **495 passed, 1 skipped** (skip is a live DeepSeek API call); `ruff` clean.

## Notes
- The bearer token carries no region, so `AWS_REGION` / `AWS_DEFAULT_REGION` is still required (falls back to `us-west-2`).
- Bedrock API keys are scoped to Bedrock + Bedrock Runtime; they do **not** cover `InvokeModelWithBidirectionalStream`, Agents, or Data Automation. TradingAgents uses Converse, which is in scope.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -25,8 +25,13 @@ NVIDIA_API_KEY=
 #OPENAI_COMPATIBLE_API_KEY=
 
 # AWS Bedrock (provider "bedrock", install with: pip install ".[bedrock]").
-# Auth uses the standard AWS credential chain; set the region (and optionally a
-# named profile). No single API key.
+# Two auth modes — pick ONE:
+#   1. Bedrock API key (simplest): a bearer token, no AWS access keys needed.
+#      AWS_BEARER_TOKEN_BEDROCK=<your-bedrock-api-key>
+#   2. AWS SigV4 credential chain: env access keys, ~/.aws/credentials, an IAM
+#      role, or a named profile (AWS_PROFILE).
+# Either way set the region (the bearer token carries none; defaults us-west-2).
+#AWS_BEARER_TOKEN_BEDROCK=
 #AWS_DEFAULT_REGION=us-west-2
 #AWS_PROFILE=
 
--- a/.env.example
+++ b/.env.example
@@ -31,6 +31,8 @@ NVIDIA_API_KEY=
 #   2. AWS SigV4 credential chain: env access keys, ~/.aws/credentials, an IAM
 #      role, or a named profile (AWS_PROFILE).
 # Either way set the region (the bearer token carries none; defaults us-west-2).
+# When the bearer token is set it WINS: any AWS_PROFILE active in your shell is
+# ignored for the Bedrock client (so a stale profile can't break the token).
 #AWS_BEARER_TOKEN_BEDROCK=
 #AWS_DEFAULT_REGION=us-west-2
 #AWS_PROFILE=
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 Breaking changes within the 0.x line are called out explicitly.
 
+## [Unreleased]
+
+### Added
+
+- **Bedrock API key auth.** Amazon Bedrock now accepts a native API key (bearer
+  token) via `AWS_BEARER_TOKEN_BEDROCK` as an alternative to the AWS SigV4
+  credential chain — no AWS access keys required. The CLI advises (never prompts)
+  when neither auth mode is configured, and the docs/`.env.example` describe both.
+
 ## [0.3.0] — 2026-06-22
 
 Stabilization and extensibility release: a CI gate, a unified verified
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,13 @@ Breaking changes within the 0.x line are called out explicitly.
 
 - **Bedrock API key auth.** Amazon Bedrock now accepts a native API key (bearer
   token) via `AWS_BEARER_TOKEN_BEDROCK` as an alternative to the AWS SigV4
-  credential chain — no AWS access keys required. The CLI advises (never prompts)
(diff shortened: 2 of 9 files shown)


praxstack:
## Update: hardened via adversarial review + verified live against Bedrock (us-east-2)

Ran a code-review → fix → re-review loop until it converged (round 2 returned **zero** findings), and exercised the feature against a **real Bedrock API key** in `us-east-2`.

### Real bug found & fixed (round 1, confirmed by live reproduction)
With a valid `AWS_BEARER_TOKEN_BEDROCK` set, a stray/invalid **ambient `AWS_PROFILE`** (or partial AWS access keys, or container-credential hints) made `ChatBedrockConverse` construction raise `ProfileNotFound` — because langchain-aws's bearer branch still builds the boto3 client through the standard session, which eagerly resolves the SigV4 chain. Since developers commonly export `AWS_PROFILE` globally, the feature's headline promise ("a bearer token alone just works") broke in common dev/CI/ECS environments.

**Fix** (`bedrock_client.py`): when a bearer token is set, construction now runs inside a context manager that transiently clears the ambient SigV4 credential env vars (lock-guarded, restored in `finally`). No-op without a token, so the SigV4 credential-chain path is untouched. The CLI advisory now also tells the user when a token overrides an active profile, and the example model IDs were corrected (the bare `-v1:0` form was malformed; real profiles look like `us.anthropic.claude-opus-4-8`).

### Verified live (us-east-2, Bedrock API key)
- Bearer-token Converse: **Opus 4.8** (incl. extended thinking), **Sonnet 4.6**, **Haiku 4.5** → PASS. `Authorization: Bearer …`, not SigV4.
- End-to-end structured-output smoke (`scripts/smoke_structured_output.py bedrock`, Opus 4.8 deep + Sonnet 4.6 quick): all 6 structure checks PASS.
- Regression proof: with a bogus `AWS_PROFILE` exported in the shell, the run that previously crashed now succeeds.

### Tests
+8 unit tests (full suite **503 passed**, ruff clean): parametrized bearer-vs-ambient-contaminant regression (fails without the fix), no-op-without-token guard for the SigV4 path, region default, and the CLI override warning.

praxstack:
## Correction + additional smoke coverage: OpenAI models on Bedrock

Correcting an earlier claim of mine: **OpenAI models *are* available on Bedrock** and authenticate cleanly through this PR's bearer-token path. Verified live in **us-east-2** via `AWS_BEARER_TOKEN_BEDROCK`:

| Model (Bedrock foundation-model ID) | Bearer-token Converse |
|---|---|
| `openai.gpt-oss-120b-1:0` | ✅ PASS |
| `openai.gpt-oss-20b-1:0` | ✅ PASS |
| `openai.gpt-oss-safeguard-20b` | ✅ PASS |
| `openai.gpt-oss-safeguard-120b` | ⚠️ authenticated; empty content (safety-classifier model quirk, not auth) |

`list_foundation_models` in us-east-2 returns providers: Amazon, Anthropic, Cohere, DeepSeek, Google, Meta, MiniMax, Mistral AI, Moonshot AI, NVIDIA, OpenAI, Qwen, Stability AI, TwelveLabs, Writer, Z.AI. There is no model literally named "codex 5.5" in the catalog; the real OpenAI models on Bedrock are the `gpt-oss` family, now smoke-tested above.

Net: the bearer-token auth in this PR works across Anthropic (Opus 4.8 / Sonnet 4.6 / Haiku 4.5) **and** OpenAI (`gpt-oss`) models on Bedrock.

Yijia-Xiao:
Thanks @praxstack. Shipped in `43bd32b`: `AWS_BEARER_TOKEN_BEDROCK` is passed to ChatBedrockConverse as `api_key`, so bearer auth wins over an ambient profile.

PR #1105, opened 2026-06-30
add evidence ledger, citation and guardrails

JerryGinger:
Introduces an evidence tracking and guardrail framework to audit, verify, and constrain LLM-generated portfolio judgments using verified market data.
- Core Modules: Added EvidenceLedger for runtime item tracking, citation extraction/checking (verify_citations), and price-target math guardrails (MathGuardrailEngine).

- PM Schema: Added supporting_evidence_ids to the structured PortfolioDecision output.

- Block Mode: Setting evidence_strict_mode=block marks failed citation or price target validations as "Blocked - Not Actionable" (re-rating to Hold), while preserving the original judgment in the report for auditing.

- Report & State Logs: Unified the main decision area to use the canonical final_trade_decision, excluded blocked decisions from performance memory logs, and persisted all strict/actionable/original metrics in evidence_audit.json.


pippo73:
This addresses a failure mode we hit repeatedly, and `verify_citations` in particular would have caught a concrete error for us: a run asserted a record inventory figure that was roughly 2% above the actual maximum in the source data, which was also trending down rather than up. Nothing in the pipeline noticed.

Two observations from building similar deterministic checks locally, offered as scope questions rather than requests:

- **Unit and currency validation appears to be outside the ledger's remit, and it is the error class that hurt us most.** For a Norwegian issuer, a model reported a 53.2 billion market cap in dollars when the figure was in kroner — a 9x error propagated coherently through the whole report, with every downstream ratio internally consistent. It is detectable without any LLM: the implied FX rate can be recovered from the payload's own arithmetic (quoted EPS against net income per share), which returned 9.221 in that case and 0.994 for a control name in euro, raising no false alarm. The same check flagged `Debt to Equity: 11.879` as a percentage rather than an 11.9:1 ratio. Would this fit as another guardrail engine alongside `MathGuardrailEngine`, or do you see it belonging in the dataflows layer instead?
- **`evidence_strict_mode=block` re-rating to Hold is the right default only if "no data" is distinguished from "neutral data".** Our sentiment sources return 404 and 429 for most tickers, so a blocked decision and a genuinely neutral one look identical in the report. Preserving the original judgment for auditing, as this PR does, mitigates it, but a reason code on the block would make the difference legible.

One note on the price-target checks: when we built an equivalent plan-consistency checker, two of its rules were wrong on first writing — anchoring thresholds to the series median rejects correct thresholds on a declining series, and ignoring direction breaks it because upgrades and downgrades have opposite rules. Worth a regression test with a declining series if you do not have one.

Tauric-Research:

PR #1115, opened 2026-07-04
fix(dataflows): actually apply the Alpha Vantage fundamentals look-ahead filter

zachthebird:
## Summary

- `_filter_reports_by_date` guards on `isinstance(result, dict)`, but `_make_api_request` returns the fundamentals payload as a JSON **string**, so the guard falls through and the filter never runs. Annual/quarterly reports dated after `curr_date` leak into historical runs, silently undoing the look-ahead protection from #475.
- Parse the JSON string before filtering and re-serialize on the way out (callers still receive a `str`). Non-JSON / non-object bodies and `curr_date=None` behave exactly as before.
- Adds a regression test in the existing `tests/test_alpha_vantage_hardening.py` style — it fails on current `main` (`['2025-12-31', '2023-12-31']` survives a `curr_date='2024-01-01'` filter) and passes with this fix. No API key required: it monkeypatches `requests.get` like the neighboring tests.

Restores the CHANGELOG guarantee for #475 ("Backtesting fetchers no longer leak look-ahead data when `curr_date` is in the middle of a fetched window"), which the current code doesn't meet for the fundamentals endpoints (`get_balance_sheet`, `get_cashflow`, `get_income_statement`).

## Test plan

```
pytest tests/test_alpha_vantage_hardening.py -q
```

4 passed with the fix; the new test fails without it.

### DIFF
--- a/tests/test_alpha_vantage_hardening.py
+++ b/tests/test_alpha_vantage_hardening.py
@@ -52,3 +52,26 @@ def test_invalid_key_not_mislabeled_as_rate_limit(monkeypatch):
     with pytest.raises(av.AlphaVantageRateLimitError):  # sanity: rate-limit path still distinct
         monkeypatch.setattr(av.requests, "get", _patched_get('{"Note": "API call frequency is 5 calls per minute."}'))
         av._make_api_request("TIME_SERIES_DAILY", {"symbol": "AAPL"})
+
+
+@pytest.mark.unit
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @zachthebird for the correct root-cause. Shipped in `3570f2e`: the fundamentals filter now parses the JSON-string payload, so future-dated reports no longer leak into historical runs.

PR #1122, opened 2026-07-08
Add a candidate screener and trade-horizon-aware analysis

cfournel:
## Summary

- Adds `scripts/screen_candidates.py`, a cheap non-interactive candidate screener (no LLM calls) that pre-filters a broad universe before a full deep-dive: Yahoo Finance's own screener for stocks (`yfinance`'s `yf.screen()`), CoinGecko's public markets endpoint for crypto, and a small static futures list for commodities — filtered by risk level (beta / market-cap rank / editorial risk tier) and ranked by momentum matching a requested trade horizon. This lets a caller find *new* candidates instead of only re-analyzing tickers it already knows about, without paying for a multi-agent run per ticker across a whole market.
- Threads an optional trade horizon (`"swing"`, a few days, vs `"position"`, a multi-month hold) through `TradingAgentsGraph.propagate()`, biasing the research manager, trader, and portfolio manager prompts toward the requested holding period via a new `get_horizon_instruction()` helper. Analysts' reports are unchanged (the underlying data doesn't change with intent), only the three decision-making stages. Horizon is folded into the checkpoint signature so a horizon change starts a fresh run instead of resuming under stale instructions.
- Adds `tradingagents.reporting.extract_screen_summary()`, which pulls a compact structured summary (direction, sentiment band/score, price target, time horizon) out of a completed run's final state — reusing fields the Portfolio Manager and Sentiment Analyst already produce as structured output (`PortfolioDecision.price_target`/`time_horizon`, the Sentiment Analyst's rendered header) rather than asking any agent for a new field.
- `scripts/batch_analyze.py` gains a `--horizon {swing,position}` flag (forwarded to `propagate()`) and now includes the extracted summary fields in its JSON output.

## Motivation

Built this to power a screener feature in a downstream integration (an [agent plugin](https://github.com/cfournel/hermes-tradingagents-plugin) that lets a scheduler/agent request TradingAgents runs), where "run the full pipeline on the whole market" isn't viable but "cheaply narrow candidates, then deep-dive the shortlist" is. Figured the screener script and horizon-awareness are generally useful independent of that integration.

## Test plan

- [x] Added unit tests for `extract_screen_summary` (structured fields present / absent) in `tests/test_reporting.py`
- [x] Extended `tests/test_checkpoint_resume.py::test_run_signature_captures_graph_shape` to assert a horizon change invalidates the checkpoint signature
- [x] Manually verified `screen_candidates.py`'s crypto path against the live CoinGecko API and its commodity fallback path
- [ ] Please run the full suite in CI — I didn't have the project's LLM/graph dependencies (`langchain_core`, `langgraph`, etc.) installed in the environment I wrote this in, so I could only compile-check and unit-test the pieces above in isolation


Tauric-Research:

PR #1125, opened 2026-07-10
Add read-only IBKR portfolio context to batch analysis

crabbag3:
## What changed

- add a read-only TWS portfolio snapshot loader using `ib_async`
- sanitize account data before it reaches prompts or saved artifacts
- inject ticker ownership, cost basis, P&L, cash, and concentration context into the Trader, risk analysts, and Portfolio Manager
- add a coordinated post-batch portfolio review with Markdown and CSV action outputs
- add `--ibkr-context` and TWS connection options to `batch-analyze`
- reconcile non-base-currency position values against account gross exposure before calculating NAV weights
- route pytest temporary/cache files to a durable writable sibling directory on Windows

## Why

Standalone ticker reports could recommend initiating positions already held, assume owned names had zero exposure, or suggest additions without considering portfolio concentration. This change gives the decision stages a frozen live account snapshot while keeping the factual research agents portfolio-blind.

## Safety and user impact

- TWS is connected with `readonly=True`
- no order placement, modification, or cancellation methods are exposed
- exactly one account is required
- account identifiers and connection secrets are excluded from prompts and artifacts
- a failed or contradictory snapshot stops the run before LLM analysis
- existing behavior is unchanged unless `--ibkr-context` is supplied

## Validation

- `python -m ruff check --ignore B008 ...` — passed
- focused IBKR/CLI/graph suite — 63 passed
- complete suite with optional live DeepSeek call disabled — 528 passed, 2 skipped, 69 subtests passed
- live TWS read-only snapshot retrieved all 12 holdings and correctly reconciled USD positions against AUD NAV


Tauric-Research:

PR #1126, opened 2026-07-10
fix(dataflows): make Yahoo news windows UTC and end-exclusiveFix/yahoo news utc window

blankerLi:
## Summary

- treat Yahoo Finance news windows as UTC half-open intervals
- exclude articles published at midnight immediately following the requested end date
- normalize nested ISO and flat Unix timestamps to UTC-aware datetimes
- preserve the existing policy for undated articles

## Root cause

The previous date-window comparison used an inclusive upper bound. For an
analysis ending on `2025-05-09`, an article published exactly at
`2025-05-10T00:00:00Z` could therefore enter the historical run.

Flat Yahoo articles were also parsed using the host machine's local timezone,
which could produce different filtering results across environments.

## Changes

- normalize naive and timezone-aware datetimes to UTC
- use a half-open `[start, end + 1 day)` news window
- parse Unix timestamps directly as UTC-aware datetimes
- normalize ISO timestamps with timezone offsets to UTC
- keep malformed or missing timestamps non-fatal
- add regression tests for boundary and timezone behavior

## Test plan

- `python -m pytest tests/test_news_lookahead.py -q` — 14 passed
- `python -m pytest -q` — 568 passed, 2 skipped
- `python -m ruff check .` — passed

No live API calls or credentials are required.

### DIFF
--- a/tests/test_news_lookahead.py
+++ b/tests/test_news_lookahead.py
@@ -5,7 +5,7 @@ Regressions for #992 (flat articles bypassed the date filter), #1007 (global
 news injected future articles), #993 (empty-after-filter returned a blank body).
 """
 import time
-from datetime import datetime
+from datetime import datetime, timezone
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


Yijia-Xiao:
Thanks @blankerLi for the precise diagnosis. Shipped in `40774ca`: the news window is now half-open with every operand normalized to UTC, so the midnight-after article can no longer leak and filtering does not depend on the host timezone.

PR #1139, opened 2026-07-12
Handle missing Windows console in CLI

### LINKED ISSUE #1138: CLI crashes with NoConsoleScreenBufferError in non-interactive Windows terminals
title=CLI crashes with NoConsoleScreenBufferError in non-interactive Windows terminals
body=### What happens
Running `tradingagents` or `python -m cli.main` in this Windows-based environment crashes before the first prompt with:

`NoConsoleScreenBufferError: No Windows console found. Are you running cmd.exe?`

The traceback comes from `questionary`/`prompt_toolkit` when it tries to initialize the first prompt.

### Why this matters
The CLI currently surfaces a stack trace instead of a short actionable message. Users running from an unsupported terminal get no guidance on how to proceed.

### Reproduction
1. In a Windows environment without a console screen buffer, run `tradingagents`.
2. The process exits with a traceback before any input can be entered.

### Expected behavior
The CLI should catch this terminal-specific error and print a concise message telling the user to run it from an interactive terminal such as Windows Terminal, PowerShell, or cmd.exe.

HUAN2022A:
## Summary
- catch `NoConsoleScreenBufferError` at the CLI entry point
- print a concise actionable message instead of a full traceback when prompt_toolkit cannot find a Windows console screen buffer
- add regression coverage for the handled terminal error while preserving normal exception behavior

Fixes #1138.

## Tests
- `uv run pytest -q tests\test_cli_console_errors.py tests\test_cli_env_skip.py tests\test_cli_config_precedence.py tests\test_cli_symbol_handling.py`
- `uv run tradingagents` in this non-interactive Windows environment now prints the short console guidance instead of a traceback

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -50,6 +50,12 @@ from tradingagents.graph.analyst_execution import (
 from tradingagents.graph.trading_graph import TradingAgentsGraph
 from tradingagents.reporting import write_report_tree
 
+try:
+    from prompt_toolkit.output.win32 import NoConsoleScreenBufferError
+except (ImportError, AttributeError):
+    class NoConsoleScreenBufferError(Exception):  # type: ignore[no-redef]
+        pass
+
 console = Console()
 
 app = typer.Typer(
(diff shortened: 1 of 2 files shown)


HUAN2022A:
Addressed the cross-platform import concern by guarding the prompt_toolkit Windows import in cli.main. The test now uses cli.main.NoConsoleScreenBufferError instead of importing the Windows-specific module directly, so non-Windows test collection does not touch prompt_toolkit.output.win32.\n\nValidation: uv run pytest -q tests\\test_cli_console_errors.py tests\\test_cli_env_skip.py tests\\test_cli_config_precedence.py tests\\test_cli_symbol_handling.py

Yijia-Xiao:
Thanks @HUAN2022A. Superseded by `3f6c082`, which gates the Windows-only import on `sys.platform` so a broken prompt_toolkit still surfaces on Windows while the handler stays inert elsewhere.

PR #1163, opened 2026-07-23
fix(dataflows): stop snapshot fundamentals leaking future data into backtests

miznan:
## Problem

`yfinance`'s `.info` and Alpha Vantage's `OVERVIEW` are **live quote snapshots**. Market cap, TTM ratios, the 52-week range and the 50/200-day averages always describe *today*, regardless of the date being analyzed.

The statement endpoints already guard against this (`filter_financials_by_date`, `_filter_reports_by_date`), but the snapshot path was missed — `y_finance.get_fundamentals` even documented its `curr_date` parameter as `"current date (not used for yfinance)"`.

So every backtest feeds the Fundamentals Analyst present-day data.

## Impact — observed, not theoretical

Running `propagate("BTC-USD", "2024-11-01")`:

| Agent | Source | 200-day average |
|---|---|---|
| Market Analyst | `get_indicators` → `close_200_sma` | **$63,385** (correct as of 2024-11-01) |
| Fundamentals Analyst | `.info` snapshot | **$72,810** (present-day value) |

The Bull argued price was 9.6% *above* its long-term trend; the Bear read the same asset as a death cross. The Research Manager then reported:

> *"200-day MA discrepancy: Bull cites $63,385; Bear cites $72,810 — this creates a fundamental disagreement on whether BTC is above or below its long-term trend line"*

...and the Portfolio Manager returned **Hold** on the grounds that the data was irreconcilable. The disagreement was entirely an artifact of the leak. The 52-week high ($126,198) and $1.31T market cap were likewise impossible values for November 2024.

This reproduced across three different models, so it is not model-specific.

## Fix

Withhold snapshot-derived fields when `curr_date` is in the past, keeping only descriptive identity fields (`Name`, `Sector`, `Industry`), and tell the agent where the as-of-date figures actually live:

```
# NOTE: 2024-11-01 is in the past and yfinance provides no point-in-time
# snapshot, so live-quote metrics (market cap, PE, 52-week range, 50/200-day
# averages) are omitted here to avoid look-ahead bias. Use
# get_income_statement / get_balance_sheet / get_cashflow for as-of-date
# financials, and get_indicators for as-of-date price statistics such as
# close_50_sma and close_200_sma.
```

- New `is_historical_date()` in `stockstats_utils.py`, alongside the existing `filter_financials_by_date`
- `y_finance.get_fundamentals` — filters fields; an empty result under a historical date no longer misreports as `NoMarketDataError`, since the data existed and was deliberately withheld
- `alpha_vantage_fundamentals` — `_strip_snapshot_fields` mirrors the shape of the existing `_filter_reports_by_date` (JSON-string in/out, non-JSON passes through). `Description` is excluded as well, since that prose is rewritten over time and can narrate events postdating `curr_date`

**Live usage is unchanged** — with `curr_date` unset or set to today, every field is returned exactly as before.

## Verification

After the fix, the same BTC-USD 2024-11-01 run contains **zero** occurrences of `72,810` / `126,198` / `$1.31T`. Both debaters reason from the single correct `close_200_sma = $63,385`, and the decision changes from the artifact-driven `Hold` to `Overweight`.

The Fundamentals Analyst handles the note gracefully — it reports that no usable statement data exists for BTC, points at `get_indicators`, and moves on.

## Test plan

- [x] `tests/test_fundamentals_snapshot_lookahead.py` — 12 cases covering both vendors: fields withheld on a historical date, fields preserved when live, the today/yesterday boundary, unparseable dates, non-JSON pass-through, `Description` exclusion, and the "everything withheld ≠ unknown symbol" case
- [x] Full suite: 587 passed. Two pre-existing failures in `test_ollama_base_url.py` are unrelated (CLI string assertions) and confirmed failing on unmodified `main`

### DIFF
--- a/tests/test_fundamentals_snapshot_lookahead.py
+++ b/tests/test_fundamentals_snapshot_lookahead.py
@@ -1,0 +1,125 @@
+"""Snapshot fundamentals must not leak the present into a backtest.
+
+Statement endpoints were already date-filtered, but the vendor "overview"
+endpoints (yfinance ``.info``, Alpha Vantage OVERVIEW) return a live quote:
+market cap, TTM ratios, the 52-week range and the 50/200-day averages always
+describe today. Replaying 2024-11-01 therefore fed the fundamentals analyst a
+200-day average from the present, contradicting the ``close_200_sma`` the market
+analyst computed as of the trade date and derailing the debate.
+"""
+import json
+
+import pandas as pd
+import pytest
+
+import tradingagents.dataflows.alpha_vantage_fundamentals as avf
+import tradingagents.dataflows.y_finance as yfin
+from tradingagents.dataflows.stockstats_utils import is_historical_date
+
+LEAKY_LABELS = ("Market Cap", "52 Week High", "50 Day Average", "200 Day Average")
+
+_INFO = {
+    "longName": "Bitcoin USD",
+    "sector": "Crypto",
+    "marketCap": 1315814113280,
+    "fiftyTwoWeekHigh": 126198.07,
+    "fiftyDayAverage": 63092.676,
+    "twoHundredDayAverage": 72809.93,
+}
+
+_OVERVIEW = json.dumps({
+    "Symbol": "AAPL",
+    "Name": "Apple Inc",
+    "Sector": "TECHNOLOGY",
+    "Description": "Apple completed its 2025 acquisition of ExampleCorp.",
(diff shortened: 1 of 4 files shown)


Yijia-Xiao:
Thanks @miznan. Fixed in 96111aa: the live company profile is withheld from historical fundamentals for both vendors.

PR #1179, opened 2026-07-27
feat: write reproducible run manifests

faceWang753:
## Summary

- write a machine-readable `run_manifest.json` alongside CLI and package API reports
- record the requested as-of date, asset type, selected analysts, provider/model IDs, configured vendor-chain IDs, a safe effective-config hash, context hashes, and the final rating/output hash
- sanitize backend URLs so credentials, query strings, and fragments are never copied into a shareable report

## Validation

- `pytest -m unit -q`: 360 passed, 1 skipped, 202 deselected (69 subtests passed)
- focused reporting tests: 5 passed
- `ruff check` on all touched Python files: passed
- `git diff --check`: passed

## Reproducibility boundary

This does **not** claim deterministic model output or exact replay. The vendor IDs describe configured chains, not proof of which fallback served a particular call; exact replay still requires archived live inputs.

One boundary question for maintainers: is configured vendor-chain identity the right first manifest boundary here, or would you prefer this to wait until per-tool served-vendor/input receipts can also be captured? Please correct this framing if it does not match the project's intended reproducibility contract.

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,15 @@ config["temperature"] = 0.0
 
 What does not vary anymore: the analyzed company identity is resolved deterministically from the ticker before any agent runs, and the market analyst grounds exact price and indicator claims in a verified data snapshot. Earlier reports of "different companies" or fabricated price levels across runs are addressed by these two mechanisms.
 
+Saved reports include `run_manifest.json`. It records the requested as-of
+date, asset type, analyst set, model/provider IDs, configured data-vendor IDs,
+the safe subset of the effective configuration and its SHA-256, hashes of the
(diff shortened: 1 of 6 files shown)


faceWang753:
The independent pinned preflight is now public:

https://github.com/faceWang753/finmirror/blob/main/audits/tradingagents/preflight-01477f9.md

It records the exact upstream revision, capability map, verification commands/results, and why no FinMirror score is claimed without frozen live-input receipts. Please correct any factual or technical boundary in the report. I will not describe this as a joint or endorsed case study without explicit maintainer permission.

PR #1189, opened 2026-07-30
fix(rating): surface unparseable ratings as REVIEW instead of a silent Hold (#1170)

### LINKED ISSUE #1170: [Bug] Rating parser silently converts unrecognized output to Hold
## Problem

The shared rating parser uses `Hold` as its default when it cannot extract a recognized five-tier rating. Because `Hold` is a valid rating, this conflates "no parseable rating" with an actual neutral decision.

The parser first looks for a canonical `Rating: X`-style label and then scans whitespace-separated tokens for the five-tier vocabulary. The fallback is sensitive to formatting details such as Unicode punctuation and localized labels. For example, output such as the following may contain a recognizable English rating but still fail exact token matching:

- `Rating：Overweight`
- `Final rating: Sell（bearish）`
- `评级：Overweight（超配）`

This request does not require the project to become Chinese-first or to support a particular output language. The agent can remain English-default. The issue is language- and format-agnostic: when free-text fallback produces a non-canonical response, an unparseable result should not be silently converted into a valid `Hold` signal.

## Impact

When structured output falls back to free text, the machine-readable rating can disagree with the human-readable Portfolio Manager decision. Downstream reports, memory entries, or signal consumers cannot distinguish a genuine `Hold` from a parser failure.

The five-tier rating and the separate three-tier trade action should also remain distinct. For example, `Overweight` is not interchangeable with `Buy` merely because both are bullish.

## Possible direction

One possible direction would be to return an explicit `REVIEW`, `None`, or equivalent parse-status value when no rating is recognized, while preserving the raw text and parsing method. The parser could normalize harmless Unicode punctuation and support explicitly documented aliases, but the exact normalization policy is up to the maintainers.

For structured-output fallback, another option is to preserve a typed rating field or attach an explicit "unparseable free-text result" status instead of relying on a valid rating as the fallback value.

## Acceptance criteria

- An unrecognized or missing rating never silently becomes `Hold`.
- Canonical English output, Markdown-wrapped output, and harmless punctuation variants are covered by tests.
- A non-canonical or localized example is either parsed according to a documented policy or surfaced as an explicit review state; multilingual prompting is not required.
- Rating and trade-action fields remain separate and are not inferred from each other without an explicit mapping.
- A mismatch between the human-readable final decision and the machine signal is visible to downstream consumers.
- No broker connection, order execution, account data, or private analysis reports are required.

This issue concerns deterministic parsing and failure handling in `tradingagents/agents/utils/rating.py`.

aniketshukla1:
## Summary

The shared rating parser (`tradingagents/agents/utils/rating.py`) returned `"Hold"` whenever it could not extract a 5-tier rating, conflating a genuine neutral call with a parse failure. It also missed ratings wrapped in fullwidth/CJK punctuation (e.g. `Rating：Overweight`, `评级：Overweight（超配）`), and its token scan could partially match longer words.

Closes #1170.

## Changes

- **`rating.py`**: add `extract_rating()`, which returns `None` when no rating is recognised (never a silent default), normalises harmless fullwidth/CJK punctuation, and matches whole alphabetic tokens (so `buyer` / `holding` / `seller` don't spuriously match). `parse_rating()` is kept as a backward-compatible wrapper. New `RATING_REVIEW` sentinel that is deliberately **not** a 5-tier value.
- **`signal_processing.py` / `memory.py`**: surface `RATING_REVIEW` for an unparseable decision, so a parse failure is visible to downstream consumers and in the decision log instead of masquerading as `Hold`. The 5-tier rating stays distinct from the 3-tier trade action.
- **tests**: add `tests/test_rating.py` (canonical / markdown / fullwidth / localized / whole-word / `None` cases); update the two existing tests that asserted the old silent-`Hold` behaviour.

## Acceptance criteria (#1170)

- [x] An unrecognized or missing rating never silently becomes `Hold`
- [x] Canonical English, Markdown-wrapped, and harmless punctuation variants are covered by tests
- [x] A localized example is parsed per a documented policy (fullwidth punctuation normalised; whole-word token matching)
- [x] Rating (5-tier) and trade-action (3-tier) fields remain separate and are not inferred from each other
- [x] A mismatch between the decision text and the machine signal is visible downstream (`REVIEW`)

### DIFF
--- a/tests/test_memory_log.py
+++ b/tests/test_memory_log.py
@@ -8,6 +8,7 @@ import pytest
 from tradingagents.agents.managers.portfolio_manager import create_portfolio_manager
 from tradingagents.agents.schemas import PortfolioDecision, PortfolioRating
 from tradingagents.agents.utils.memory import TradingMemoryLog
+from tradingagents.agents.utils.rating import RATING_REVIEW
 from tradingagents.graph.propagation import Propagator
 from tradingagents.graph.reflection import Reflector
 from tradingagents.graph.trading_graph import TradingAgentsGraph
@@ -171,10 +172,12 @@ class TestTradingMemoryLogCore:
         log.store_decision("AAPL", "2026-01-11", DECISION_OVERWEIGHT)
         assert log.load_entries()[0]["rating"] == "Overweight"
 
-    def test_rating_fallback_hold(self, tmp_path):
+    def test_rating_review_when_unparseable(self, tmp_path):
+        # A decision with no parseable rating is tagged REVIEW, not a silent Hold,
+        # so the log never conflates a parse failure with a genuine neutral call.
         log = make_log(tmp_path)
         log.store_decision("MSFT", "2026-01-12", DECISION_NO_RATING)
-        assert log.load_entries()[0]["rating"] == "Hold"
+        assert log.load_entries()[0]["rating"] == RATING_REVIEW
 
     def test_rating_priority_over_prose(self, tmp_path):
         """'Rating: X' label wins even when an opposing rating word appears earlier in prose."""
--- a/tests/test_rating.py
+++ b/tests/test_rating.py
@@ -1,0 +1,101 @@
+"""Tests for the deterministic rating parser in ``tradingagents.agents.utils.rating``.
+
+Covers the contract from issue #1170: an unrecognised or missing rating must never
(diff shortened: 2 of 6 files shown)


Yijia-Xiao:
Thanks @aniketshukla1. Superseded by `43fc275`: unparseable ratings surface REVIEW via extract_rating()/is_review, with parse_rating kept for compatibility callers.

PR #1195, opened 2026-08-01
feat(llm): add openai_codex provider using ChatGPT subscription quota

### LINKED ISSUE #1194: feat: bill runs to a ChatGPT subscription via an openai_codex provider
## Problem

A full run fans out across analysts, two debates and a portfolio manager, so it burns a fair number of tokens. Users who already pay for a ChatGPT subscription currently cannot put that quota to work here — every OpenAI-family path in the repo bills API credits.

## Why `openai_compatible` does not cover it

The natural first attempt is `llm_provider: "openai_compatible"` with `backend_url` pointed at the Codex endpoint. That fails, because the endpoint breaks three assumptions the shared OpenAI client makes:

1. **Auth is not an API key.** Credentials live in the OAuth token the official Codex app/CLI writes to `~/.codex/auth.json`. The token also has to be refreshed near expiry, and the refresh token *rotates* — writing the new one back is required, or the Codex client itself gets logged out.
2. **`stream: false` and `store: true` are rejected outright**, and `temperature` is not accepted.
3. **Input items with role `system` return 400.** Every agent prompt in this repo is a `ChatPromptTemplate` whose first message is a system message, so the first analyst node fails on its first call and the run stalls before any agent completes. The fix is to relabel those items `developer` (the Responses-API name for the same role) on the way out; adding `instructions` does not make `system` acceptable.

Points 1 and 3 in particular cannot be expressed through config — they need a provider entry and a small `ChatOpenAI` subclass.

## Proposal

Add an `openai_codex` provider to the existing registry:

- `ProviderSpec.credentials_fn` resolves (and refreshes/rotates) the token from `~/.codex/auth.json`, overridable with `TRADINGAGENTS_CODEX_AUTH_PATH`.
- A `CodexChatOpenAI` subclass pins the payload constraints above and relabels system messages.
- CLI picker entry plus a login preflight, so an unauthenticated user gets "run `codex login` first" instead of a 401 mid-run.
- No behaviour changes anywhere unless the provider is selected.

## Caveat worth your call

This endpoint is undocumented and unversioned — OpenAI can change it without notice — and driving subscription credentials from a third-party tool is not clearly sanctioned by OpenAI's terms. I documented both points in the README next to the provider so users choose with open eyes, but whether TradingAgents wants to ship it at all is a maintainer decision, and I would rather ask than assume.

If you would prefer this to live outside the repo, or to be gated behind an extra opt-in, I am happy to rework it.

## Status

Implementation is ready and linked below: 15 files, `pytest -q` at 614 passed / 5 skipped, `ruff check .` clean.

LeonWTW:
Closes #1194 — that issue carries the full rationale; this is the implementation.

Adds an `openai_codex` provider so a run can be billed to an existing ChatGPT subscription instead of API credits. Nothing changes for any other provider: every code path here is behind the new provider name.

## Why this needs a provider rather than `openai_compatible` + `backend_url`

The endpoint breaks three assumptions the shared OpenAI client makes:

| Assumption | Reality | Handled in |
|---|---|---|
| Auth comes from an env var | OAuth token in `~/.codex/auth.json`, refreshed near expiry, refresh token **rotates** and must be written back or the Codex client logs out | `llm_clients/codex_auth.py` via `ProviderSpec.credentials_fn` |
| `stream: false` / `store: true` / `temperature` are fine | All rejected | `CodexChatOpenAI` — first two pinned as **constructor** kwargs (a class-level field default sets the payload but not langchain's streaming dispatch), `temperature` filtered in `get_llm` |
| `system` role is fine | 400 "System messages are not allowed" — and every agent prompt here is a `ChatPromptTemplate` starting with a system message, so the first analyst node fails on its first call | `_get_request_payload` relabels those items `developer`, the Responses-API name for the same role |

## What changed

- `tradingagents/llm_clients/codex_auth.py` (new) — resolve, refresh, rotate, persist. Path overridable via `TRADINGAGENTS_CODEX_AUTH_PATH`.
- `tradingagents/llm_clients/openai_client.py` — `CodexChatOpenAI`; auth-error messages stay actionable after the file re-read.
- `tradingagents/llm_clients/model_catalog.py`, `api_key_env.py` — registry entry; no API key is read for this provider.
- `tradingagents/graph/trading_graph.py` — `_get_provider_kwargs` forwards `openai_reasoning_effort` for `openai_codex` as well.
- `cli/utils.py`, `cli/main.py` — picker entry, login preflight (so an unauthenticated user gets "sign in with the Codex client first" instead of a 401 mid-run), and the reasoning-effort prompt.
- `README.md` — provider docs, including the caveats below.

## Trying it

```bash
codex login          # or sign in with the Codex desktop app
tradingagents        # pick "OpenAI Codex (ChatGPT subscription quota)"
```

Subscription traffic hits transient "servers are currently overloaded" errors far more often than the paid API, so the README suggests `TRADINGAGENTS_LLM_MAX_RETRIES=5` or higher.

## Testing

`pytest -q` → **614 passed, 5 skipped**; `ruff check .` → clean.

New coverage: auth-file resolution/refresh/rotation (`test_codex_auth.py`), provider registration and payload shaping (`test_codex_provider.py`), CLI preflight (`test_codex_cli.py`). `test_codex_smoke.py` drives the live endpoint with a real `ChatPromptTemplate` and is opt-in behind `-m smoke`, so CI never calls out. The smoke test matters for the system→developer fix specifically: a bare-string `invoke` carries no system message and passed throughout, which is how that bug survived.

## Caveats

This endpoint is undocumented and unversioned, so OpenAI can change it without notice, and driving subscription credentials from a third-party tool is not clearly sanctioned by OpenAI's terms. Both points are stated in the README next to the provider. If you would rather gate this further or not carry it at all, say so in #1194 and I will rework or withdraw it.

https://claude.ai/code/session_01FFuMQCxPhTkWLwrCQ57tTp

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,8 @@ For local models, configure Ollama with `llm_provider: "ollama"`. The default en
 
 For any other OpenAI-compatible server (vLLM, LM Studio, llama.cpp, or a custom relay), use `llm_provider: "openai_compatible"` and set the endpoint via `backend_url` (or `TRADINGAGENTS_LLM_BACKEND_URL`), e.g. `http://localhost:8000/v1` for vLLM or `http://localhost:1234/v1` for LM Studio. The model is whatever your server serves. No key is needed for local servers; set `OPENAI_COMPATIBLE_API_KEY` when the endpoint requires one.
 
+To bill runs to a ChatGPT subscription instead of API credits, use `llm_provider: "openai_codex"`. It reuses the login held by the official Codex app or CLI (`~/.codex/auth.json`, overridable with `TRADINGAGENTS_CODEX_AUTH_PATH`), so sign in there first; no API key is needed and none is read. TradingAgents refreshes and rewrites that file when the token nears expiry, because the refresh token rotates and not persisting it would log the Codex client out. Set `TRADINGAGENTS_LLM_MAX_RETRIES=5` or higher: subscription traffic hits transient "servers are currently overloaded" errors far more often than the paid API. Note that this endpoint is undocumented and unversioned — OpenAI can change it without notice — and that driving subscription credentials from a third-party tool is not clearly sanctioned by OpenAI's terms.
+
 Alternatively, copy `.env.example` to `.env` and fill in your keys:
 ```bash
 cp .env.example .env
@@ -197,7 +199,7 @@ An interface will appear showing results as they load, letting you track the age
 
 ### Implementation Details
 
-We built TradingAgents with LangGraph to ensure flexibility and modularity. The framework supports multiple LLM providers: OpenAI, Google, Anthropic, xAI, DeepSeek, Qwen (Alibaba DashScope, international and China endpoints), GLM (Zhipu), MiniMax (global + China), OpenRouter, Ollama for local models, and Azure OpenAI for enterprise.
+We built TradingAgents with LangGraph to ensure flexibility and modularity. The framework supports multiple LLM providers: OpenAI, OpenAI Codex (ChatGPT subscription), Google, Anthropic, xAI, DeepSeek, Qwen (Alibaba DashScope, international and China endpoints), GLM (Zhipu), MiniMax (global + China), OpenRouter, Ollama for local models, and Azure OpenAI for enterprise.
 
 ### Python Usage
 
--- a/cli/main.py
+++ b/cli/main.py
@@ -711,7 +711,10 @@ def get_user_selections():
             "Gemini thinking mode", "Step 8: Thinking Mode",
             "Configure Gemini thinking mode", ask_gemini_thinking_config,
         )
-    elif provider_lower == "openai":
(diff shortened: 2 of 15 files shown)


LeonWTW:
CI is held at `action_required` here pending maintainer approval, so I ran the same workflow on my fork in the meantime — all six jobs green (py3.10–3.13, `ruff` strict full-repo, clean-install smoke):

https://github.com/LeonWTW/TradingAgents/actions/runs/30714455163

No rush on my side; flagging it only so the approval decision doesn't have to double as a "does it even build" question.

Yijia-Xiao:
Thanks for the work here. We are not shipping this while it depends on an undocumented endpoint and on reading and rewriting the Codex client's credential file, which the provider has not sanctioned. Keeping it open: if OpenAI documents subscription-backed API access, we will add official support for it.

PR #1200, opened 2026-08-04
fix: prevent fabricated claims in opening debate turns

### LINKED ISSUE #1176: [Bug] First debate speaker fabricates the opponent's argument (`current_response` starts empty)
# [Bug] First debate speaker fabricates the opponent's argument (`current_response` starts empty)

## Summary

In both the investment debate (`Bull Researcher` / `Bear Researcher`) and the risk debate (`Aggressive` / `Conservative` / `Neutral` Analyst), the graph fixes a strict speaking order, and the agent that speaks first is given an empty "last argument from the other side" (`current_response` / `current_conservative_response` / `current_neutral_response` all start as `""`). On a local run (AAPL, 2026-07-22), the first speaker routinely writes as though it is rebutting a specific opposing argument that was never made. This reproduces regardless of which side the graph puts first, and persists even where the prompt already contains an explicit instruction telling the agent what to do when the other side hasn't spoken yet.

## Code evidence

`investment_debate_state` is initialized with `current_response` as an empty string:

```python
# TradingAgents/tradingagents/graph/propagation.py, create_initial_state()
"investment_debate_state": InvestDebateState(
    {
        "bull_history": "",
        "bear_history": "",
        "history": "",
        "current_response": "",   # L46
        "judge_decision": "",
        "count": 0,
    }
),
```

The graph fixes a strict speaking order — `Bull Researcher` always runs before `Bear Researcher`:

```python
# TradingAgents/tradingagents/graph/setup.py, L135
workflow.add_edge(current_clear, "Bull Researcher")
```

`bull_researcher.py` and `bear_researcher.py` both interpolate `current_response` directly into the prompt as "the other side's last argument," with no guard for the empty case:

```python
# TradingAgents/tradingagents/agents/researchers/bull_researcher.py, L43
Conversation history of the debate: {history}
Last bear argument: {current_response}
```

```python
# TradingAgents/tradingagents/agents/researchers/bear_researcher.py, L45
Conversation history of the debate: {history}
Last bull argument: {current_response}
```

When `count == 0`, whichever of these runs first receives `current_response = ""` and is still instructed ("Engagement: ... engaging directly with the [other]'s points") to rebut it.

**Reproduction: same pattern in both speaking orders**

- Default order (Bull → Bear), smoke test on AAPL 2026-07-22 (`logs/debate_full.md`): the Bull Analyst's *first and only* line opens with

  > "Look, I hear what you're saying about the 'higher-for-longer' interest rate environment. It's the standard playbook for bears: point to the Fed, point to the discount rates, and try to paint a picture of a stagnant ceiling..."

  — even though no Bear argument exists yet at that point in the run (`current_response` was `""`).

- Reversed order (Bear → Bull), produced by editing `setup.py` L135 to `workflow.add_edge(current_clear, "Bear Researcher")`, rerunning, then reverting (`logs/debate_reversed.md`): the Bear Analyst, now speaking first with an empty `current_response`, opens with

  > "Look, I understand why the bulls are cheering right now — the charts look 'clean,' and Apple is still the darling of the tech world..."

  — again rebutting a Bull argument that has not been produced yet.

The same pattern (opening by rebutting a specific claim from the side that hasn't spoken) was also observed by direct inspection in the Bull-first transcripts of `expC_C1_iter2.json`, `iter5.json`, `iter9.json`, and `iter13.json` (not an exhaustive audit of all 20 C1 runs, but every transcript sampled showed it).

**The prompt-level avoidance instruction exists elsewhere in the codebase and is demonstrably ineffective**

The risk-debate agents (`aggressive_debator.py`, `conservative_debator.py`, `neutral_debator.py`) — unlike the bull/bear researchers — *do* have an explicit guard for the empty case:

```python
# TradingAgents/tradingagents/agents/risk_mgmt/aggressive_debator.py, L35
Here is the current conversation history: {history} Here are the last arguments from
the conservative analyst: {current_conservative_response} Here are the last arguments
from the neutral analyst: {current_neutral_response}. If there are no responses from
the other viewpoints yet, present your own argument based on the available data.
```

`Aggressive Analyst` is structurally the first risk-debate speaker (`setup.py` L145: `workflow.add_edge("Trader", "Aggressive Analyst")`), so `current_conservative_response` and `current_neutral_response` are both `""` when this instruction fires. Despite the explicit instruction, the observed output still fabricates both other analysts' positions (`logs/risk_debate_excerpt.txt`):

> "Aggressive Analyst: Look, I understand the hesitation from both sides — the conservative analyst is terrified of a 'valuation trap' because of these high interest rates, and the neutral analyst wants to wait for some kind of 'clarity' before moving..."

At the time this text was generated, `current_conservative_response` and `current_neutral_response` were both empty strings — the specific positions attributed to the Conservative and Neutral analysts ("terrified of a valuation trap," "wants to wait for clarity") were invented, not read from state. This shows the fabrication is not simply a missing-instruction problem: an explicit "don't assume, argue from data if no response yet" instruction is present in the prompt and does not prevent it, at least not reliably, for this model.

## Reproduction

**Environment**
- Repo: `TauricResearch/TradingAgents`, commit `a33fd4c0f134485a43553a2c23a63cb14adbd88f` (2026-07-18, v0.3.1)
- Python 3.12.8, Windows
- `TRADINGAGENTS_LLM_PROVIDER=openai_compatible`
- `TRADINGAGENTS_DEEP_THINK_LLM=google/gemma-4-12b-qat`
- `TRADINGAGENTS_QUICK_THINK_LLM=google/gemma-4-12b-qat` (same model for both roles)
- `TRADINGAGENTS_LLM_BACKEND_URL=http://localhost:1234/v1` (local LM Studio)
- `TRADINGAGENTS_MAX_DEBATE_ROUNDS=1`, `TRADINGAGENTS_MAX_RISK_ROUNDS=1`
- Temperature: unset (provider/model default — not fixed to 0, so run-to-run wording varies)
- `selected_analysts=("market", "news")`, ticker `AAPL`, trade date `2026-07-22`

**Steps**
1. Default-order run: `python smoke_test.py` (repo root) — produces the full graph output including `investment_debate_state`; inspect `bull_history`/`bear_history`/`current_response` in the resulting state dump. → `logs/debate_full.md`, `logs/full_dump_raw.txt`.
2. Reversed-order run: edit `TradingAgents/tradingagents/graph/setup.py` L135 from `workflow.add_edge(current_clear, "Bull Researcher")` to `workflow.add_edge(current_clear, "Bear Researcher")`, rerun step 1, then revert the edit. → `logs/debate_reversed.md`.
3. Risk-debate transcript: same smoke-test run as step 1; inspect `risk_debate_state` (`aggressive_history`, `current_conservative_response`, `current_neutral_response`). → `logs/risk_debate_excerpt.txt`.

## Limitations

- **Single ticker, single date**: all results are from AAPL on 2026-07-22 only. No other ticker, sector, or market regime was tested.
- **Single model**: both `deep_think_llm` and `quick_think_llm` were the same local quantized model (`google/gemma-4-12b-qat`) served via LM Studio. Behavior of the hosted models the project defaults to (e.g. GPT-5.x) was not tested and may differ substantially.
- **Temperature not fixed**: sampling temperature was left at the provider/model default rather than pinned (e.g. to 0), so some run-to-run variance is expected.
- **Order-reversal test ran once per direction**: the controlled speaking-order swap (`setup.py` L135) was run once for the default order and once for the reversed order, not repeated n times; the additional four C1 transcripts checked (iter2/5/9/13) used the default order only, so the reversed-order observation itself is a single data point corroborated qualitatively rather than statistically.
- **Fabrication was assessed qualitatively**: identified by manually reading transcripts against the known-empty state fields, not by an automated large-n fabrication detector, so no fabrication *rate* is reported — only that the pattern was observed in every transcript checked.
- **No claim about trading outcomes**: this issue is about prompt/state-handling behavior (what the agent asserts about a nonexistent prior turn), not about whether resulting ratings were profitable or otherwise correct in hindsight.

This should be read as a reproducible local observation on one configuration, not as a general claim about the framework's behavior across models, tickers, or dates.

daleselaji-dev:
## Summary
- avoid presenting an empty debate state as a concrete opposing argument
- make opening bull and bear turns explicitly evidence-led
- add regression coverage for both opening prompts

## Problem
The first researcher receives an empty `current_response`, but the prompt labels it as the other side's last argument. Models can therefore rebut claims that were never made, which makes the debate history less trustworthy.

## Solution
Use a role-specific opening-context sentence when `current_response` is empty. Later turns continue to receive the actual previous argument unchanged.

## Tests
- `python -m pytest tests/test_debate_opening_prompt.py -q` (not run: pytest is not installed in the local runtime)
- `python -m compileall` (source syntax checked)

## Scope and risk
Prompt construction and two focused tests only. No graph ordering, state schema, or trading logic changes.

Closes #1176

### DIFF
--- a/tradingagents/agents/researchers/bear_researcher.py
+++ b/tradingagents/agents/researchers/bear_researcher.py
@@ -11,6 +11,7 @@ def create_bear_researcher(llm):
         bear_history = investment_debate_state.get("bear_history", "")
 
         current_response = investment_debate_state.get("current_response", "")
+        last_bull_argument = current_response or "No bull argument has been presented yet. Build the opening bear case from the available evidence without attributing claims to the bull side."
         market_research_report = state["market_report"]
         sentiment_report = state["sentiment_report"]
         news_report = state["news_report"]
@@ -42,7 +43,7 @@ Social media sentiment report: {sentiment_report}
 Latest world affairs news: {news_report}
 {fundamentals_label}: {fundamentals_report}
 Conversation history of the debate: {history}
-Last bull argument: {current_response}
+Last bull argument: {last_bull_argument}
 Use this information to deliver a compelling bear argument, refute the bull's claims, and engage in a dynamic debate that demonstrates the risks and weaknesses of investing in the {target_label}.
 """ + get_language_instruction()
 
--- a/tradingagents/agents/researchers/bear_researcher.py
+++ b/tradingagents/agents/researchers/bear_researcher.py
@@ -10,8 +10,12 @@ def create_bear_researcher(llm):
         history = investment_debate_state.get("history", "")
         bear_history = investment_debate_state.get("bear_history", "")
 
-        current_response = investment_debate_state.get("current_response", "")
-        last_bull_argument = current_response or "No bull argument has been presented yet. Build the opening bear case from the available evidence without attributing claims to the bull side."
+        initial_bull_arg = (
+            "No bull argument has been presented yet. Build the opening bear case from the "
+            "available evidence without attributing claims to the bull side."
+        )
+        current_response = investment_debate_state.get("current_response", initial_bull_arg)
+        last_bull_argument = current_response
         market_research_report = state["market_report"]
         sentiment_report = state["sentiment_report"]
         news_report = state["news_report"]
--- a/tests/test_debate_opening_prompt.py
+++ b/tests/test_debate_opening_prompt.py
@@ -1,0 +1,40 @@
+from unittest.mock import MagicMock
+
+from tradingagents.agents.researchers.bear_researcher import create_bear_researcher
+from tradingagents.agents.researchers.bull_researcher import create_bull_researcher
+
+
+def _state():
+    return {
+        "investment_debate_state": {
+            "history": "",
+            "bull_history": "",
+            "bear_history": "",
+            "current_response": "",
+            "count": 0,
+        },
+        "market_report": "market",
+        "sentiment_report": "sentiment",
+        "news_report": "news",
+        "fundamentals_report": "fundamentals",
+    }
+
+
+def test_bull_opening_does_not_attribute_missing_bear_argument():
+    llm = MagicMock()
+    llm.invoke.return_value.content = "opening case"
+
+    create_bull_researcher(llm)(_state())
+
+    prompt = llm.invoke.call_args.args[0]
+    assert "No bear argument has been presented yet" in prompt
+
+
+def test_bear_opening_does_not_attribute_missing_bull_argument():
+    llm = MagicMock()
+    llm.invoke.return_value.content = "opening case"
+
+    create_bear_researcher(llm)(_state())
+
+    prompt = llm.invoke.call_args.args[0]
+    assert "No bull argument has been presented yet" in prompt
--- a/tradingagents/agents/researchers/bull_researcher.py
+++ b/tradingagents/agents/researchers/bull_researcher.py
@@ -11,6 +11,7 @@ def create_bull_researcher(llm):
         bull_history = investment_debate_state.get("bull_history", "")
 
         current_response = investment_debate_state.get("current_response", "")
+        last_bear_argument = current_response or "No bear argument has been presented yet. Build the opening bull case from the available evidence without attributing claims to the bear side."
         market_research_report = state["market_report"]
         sentiment_report = state["sentiment_report"]
         news_report = state["news_report"]
@@ -40,7 +41,7 @@ Social media sentiment report: {sentiment_report}
 Latest world affairs news: {news_report}
 {fundamentals_label}: {fundamentals_report}
 Conversation history of the debate: {history}
-Last bear argument: {current_response}
+Last bear argument: {last_bear_argument}
 Use this information to deliver a compelling bull argument, refute the bear's concerns, and engage in a dynamic debate that demonstrates the strengths of the bull position.
 """ + get_language_instruction()
 


tradingagents/agents/researchers/bear_researcher.py
```diff
@@ -11,6 +11,7 @@ def bear_node(state) -> dict:
         bear_history = investment_debate_state.get("bear_history", "")
 
         current_response = investment_debate_state.get("current_response", "")
+        last_bull_argument = current_response or "No bull argument has been presented yet. Build the opening bear case from the available evidence without attributing claims to the bull side."
```
sjiekak: I would say 

```python
initial_bull_arg="No bull argument has been presented yet. Build the opening bear case from the available evidence without attributing claims to the bull side."
current_response = investment_debate_state.get("current_response", initial_bull_arg)
```

taro0915:
Two things from the #1176 side that might be worth folding in.

The bear researcher reaches the same empty-state path. Not under the current wiring, where the bull opens, but under a reversed debate order or a state constructed without a prior bull turn. Guarding only the opener ties the fix to the graph wiring.

It may also be worth being explicit in the description that the prompt-level cue is not sufficient on its own. The risk debators already carry an equivalent instruction ("If there are no responses from the other viewpoints yet, present your own argument based on the available data") and still fabricated both other analysts' positions in my run, details in #1176. This removes the empty-labelled-field trigger rather than fabrication in general, and saying so would keep it from being read as closing the broader problem.

I have the bear-side change and a rendered-prompt test in #1210 if either is useful here.

Yijia-Xiao:
Thanks @daleselaji-dev. Superseded by `539eae8`, which covers the risk debators too, not just bull/bear.

PR #1205, opened 2026-08-06
fix(llm): cap max_tokens and enable streaming to prevent hangs with post-7/31 deepseek-v4-flash (#1204)

### LINKED ISSUE #1204: Bug: TradingAgents hangs with post-7/31 deepseek-v4-flash — long reasoning chains exhaust token budget / trip gateway idle timeout
### Environment
- Branch: `dev` (15caeb9, 2026-08-02)
- Provider: `openai` (OpenAI-compatible) with `TRADINGAGENTS_LLM_BACKEND_URL=https://opencode.ai/zen/go/v1`
- Models: `deepseek-v4-flash` (both quick & deep think; the 7/31 stable release)
- Ticker: `159566.SZ` (A-share ETF), all 4 analysts + full pipeline

### Symptom
The run hangs indefinitely during the analyst/data-collection phase: process sits in `wait_woken` with near-zero CPU, the log stops updating after the initial data-fetch errors (StockTwits/Reddit/Polymarket are all non-fatal, expected for A-shares), and no report directory is ever created. Reproduced 3/3 times (pro+flash, flash-only, strict .env flash). A similar run on 2026-07-30 with the previous (pre-7/31) v4-flash completed fine.

### Root-cause evidence (direct API tests against the same backend)
| Prompt | max_tokens | Result |
|--------|-----------|--------|
| trivial ("reply OK") | 100 | OK — content returned, reasoning = 26 tokens |
| complex (analysis request) | 500 | **empty content** — `reasoning_tokens: 500`, `finish_reason: length` (reasoning chain ate the whole budget) |
| complex | 2048 | OK — reasoning 529 + content 55 tokens |
| complex | **not set** | **no response in 90s** (curl HTTP 000 timeout) |

### Analysis
1. The 7/31 stable v4-flash (post-training) produces much longer reasoning chains — 500+ tokens for complex tasks vs ~26 for trivial ones.
2. `_PASSTHROUGH_KWARGS` in `llm_clients/openai_client.py` does **not** include `max_tokens`, so TradingAgents sends no upper bound.
3. With no bound, the backend lets the reasoning chain run long, and a single generation exceeds the ~3 min idle timeout of the gateway (the same timeout documented in #782 for opencode-go) → connection dropped → process hangs.
4. With a small bound (wherever a default applies), the reasoning chain consumes the entire budget and `content` comes back empty → downstream agents get nothing to parse.

### Proposed fix
1. Inject a sensible default `max_tokens` (e.g. 8192) into the OpenAI-compatible client calls, overridable via a new `TRADINGAGENTS_MAX_TOKENS` env var (mirroring the existing `_ENV_OVERRIDES` pattern).
2. Consider enabling streaming for the OpenAI-compatible path (as #782 did for opencode-go), including `reasoning_content` capture from stream deltas, so long generations keep the socket alive.

### Related
- #782 (opencode-go provider: streaming required, gateway 3-min idle timeout, reasoning_content delta capture)
- #1199 (DeepSeek thinking models + tool_choice → very slow responses, adjacent class of problem)

ariesy:
## Summary

Fixes #1204. TradingAgents hangs indefinitely with the post-7/31 stable `deepseek-v4-flash` (long reasoning chains). Root cause is two-fold:

1. **No `max_tokens` sent.** `_PASSTHROUGH_KWARGS` in `llm_clients/openai_client.py` does not forward `max_tokens`. With the new long-reasoning v4-flash, an unbounded generation exceeds the gateway's ~3 min idle timeout and the connection is dropped — the process sits in `wait_woken` forever. When a bounded default does apply, the reasoning chain can consume the entire budget and `content` comes back empty.
2. **No streaming for OpenAI-compatible backends.** The OpenCode Go gateway (and similar proxies) closes connections on long non-streamed responses; streaming keeps the socket active.

Verified against `opencode.ai/zen/go/v1` with `deepseek-v4-flash`:

| Prompt | max_tokens | Result |
|--------|-----------|--------|
| trivial | 100 | OK (reasoning 26 tokens) |
| complex | 500 | **empty content** (reasoning ate all 500) |
| complex | 2048 | OK (reasoning 529 + content 55) |
| complex | unset | **no response in 90s** (HTTP 000 timeout) |

## Changes

- **`tradingagents/default_config.py`**: add `TRADINGAGENTS_MAX_TOKENS` env override + `max_tokens: 8192` default.
- **`tradingagents/graph/trading_graph.py`**: forward `max_tokens` from config into LLM client kwargs.
- **`tradingagents/llm_clients/openai_client.py`**:
  - add `max_tokens` to `_PASSTHROUGH_KWARGS`;
  - enable `streaming` when `base_url` points at `opencode.ai` (keeps the socket alive; langchain aggregates chunks back to a single AIMessage);
  - use `DeepSeekChatOpenAI` for opencode.ai + deepseek models so `reasoning_content` is captured on the streaming path via `_convert_chunk_to_generation_chunk` and round-tripped on the next turn.

## Verification

- Unit tests: 94 passed (incl. `test_deepseek_reasoning`, `test_capabilities`, `test_env_overrides`, `test_openai_compatible_provider`, `test_llm_max_retries`, `test_temperature_config`).
- End-to-end: full 159566.SZ run (all 4 analysts + research + trader + risk debate) completes; previously hung 3/3 times at the analyst phase.
- `max_tokens` default (8192) and env override (`TRADINGAGENTS_MAX_TOKENS=16384`) both verified.

## Related

- #782 (opencode-go provider: streaming required, gateway 3-min idle timeout, reasoning_content delta capture)
- #1199 (DeepSeek thinking models + tool_choice → very slow responses)

### DIFF
--- a/tradingagents/default_config.py
+++ b/tradingagents/default_config.py
@@ -19,6 +19,7 @@ _ENV_OVERRIDES = {
     "TRADINGAGENTS_BENCHMARK_TICKER":     "benchmark_ticker",
     "TRADINGAGENTS_TEMPERATURE":          "temperature",
     "TRADINGAGENTS_LLM_MAX_RETRIES":      "llm_max_retries",
+    "TRADINGAGENTS_MAX_TOKENS":           "max_tokens",
     # Provider-specific reasoning/thinking knobs (None = each provider's own
     # default). Settable here for non-interactive runs; the CLI also offers an
     # interactive choice, which is skipped when the matching var is set.
@@ -100,6 +101,12 @@ DEFAULT_CONFIG = _apply_env_overrides({
     # provider/SDK at its own default (usually 2). Raise it to ride out bursty
     # 429 throttling on rate-limited deployments instead of aborting a run (#1091).
     "llm_max_retries": None,
+    # Max output tokens per LLM call. DeepSeek V4 thinking models emit long
+    # reasoning chains; without an explicit cap the backend can stream the
+    # chain indefinitely (gateway idle timeout) or truncate content to empty
+    # (#1204). Default 8192 leaves room for reasoning + content; override via
+    # TRADINGAGENTS_MAX_TOKENS.
+    "max_tokens": 8192,
     # Checkpoint/resume: when True, LangGraph saves state after each node
     # so a crashed run can resume from the last successful step.
     "checkpoint_enabled": False,
--- a/tradingagents/graph/trading_graph.py
+++ b/tradingagents/graph/trading_graph.py
@@ -183,6 +183,13 @@ class TradingAgentsGraph:
         if max_retries is not None and max_retries != "":
             kwargs["max_retries"] = _coerce_max_retries(max_retries)
 
+        # Max output tokens per LLM call. DeepSeek V4 thinking models emit long
+        # reasoning chains; an explicit cap keeps content from being truncated to
+        # empty or the gateway from idling out on an unbounded chain (#1204).
+        max_tokens = self.config.get("max_tokens")
+        if max_tokens is not None and max_tokens != "":
+            kwargs["max_tokens"] = int(max_tokens)
+
         return kwargs
 
     def _create_tool_nodes(self) -> dict[str, ToolNode]:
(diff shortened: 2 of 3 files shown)


Yijia-Xiao:
Thanks @ariesy. Shipped the output-token cap in `0ef56e6` (opt-in, forwarded per provider); we left out the streaming change to keep the knob minimal.

PR #1206, opened 2026-08-06
harden: disable external XML entity processing in reddit.py...

anupamme:
## Summary
Harden input handling in `tradingagents/dataflows/reddit.py` (flagged by semgrep).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | gitlab.bandit.B313.B314.B315.B316.B318.B319.B320.B405.B406.B407.B408.B409.B410 |
| **Severity** | HIGH |
| **Scanner** | semgrep |
| **Rule** | `gitlab.bandit.B313.B314.B315.B316.B318.B319.B320.B405.B406.B407.B408.B409.B410` |
| **File** | `tradingagents/dataflows/reddit.py:26` |
| **Assessment** | Defensive hardening |

**Description**: Found use of the native Python XML libraries, which is vulnerable to XML external entity (XXE)
attacks. The Python documentation recommends the 'defusedxml' library instead. Use 'defusedxml'.
See https://github.com/tiran/defusedxml for more information.

## Threat Model Context

This is a CLI tool - exploitation requires the attacker to control arguments or input files passed to the tool.

## Changes
- `tradingagents/dataflows/reddit.py`
- `pyproject.toml`

## Behavior Preservation
The change is scoped to 2 files on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,7 @@ dependencies = [
     "pytz>=2025.2",
     "questionary>=2.1.0",
     "redis>=6.2.0",
+    "defusedxml>=0.7.1",
     "requests>=2.32.4",
     "rich>=14.0.0",
     "typer>=0.21.0",
--- a/tests/test_reddit_fallback.py
+++ b/tests/test_reddit_fallback.py
@@ -192,6 +192,31 @@ class TestFormatterHandlesRssPosts:
         assert "via RSS" not in out
 
 
+_XXE_ATOM = """\
+<?xml version="1.0"?>
+<!DOCTYPE feed [<!ENTITY xxe SYSTEM "file:///etc/passwd">]>
+<feed xmlns="http://www.w3.org/2005/Atom">
+  <entry><title>&xxe;</title></entry>
+</feed>
+"""
+
+
+@pytest.mark.unit
+class TestXxeSecurity:
+    """defusedxml must block XXE payloads without crashing the pipeline."""
+
+    def test_xxe_payload_rejected_gracefully(self):
+        with patch.object(reddit, "urlopen", return_value=_resp(lambda: _XXE_ATOM.encode())):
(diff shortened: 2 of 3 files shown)


yunaremaia:
Implemented in #1218: the Reddit Atom parser now uses defusedxml.ElementTree (forbid_entities/forbid_external on by default), catches EntitiesForbidden so an XXE feed degrades to [] instead of raising, and ships 3 new unit tests. Full suite: 606 passed. This also resolves the semgrep bandit.B313–B320 finding on reddit.py:26.

anupamme:
Thanks for the feedback. I agree the current diff is broader than it needs to be.

I’ll rebase #1206 onto the latest main and make this a focused security fix: replace the unsafe XML parser, add the dependency, and add regression tests covering XXE rejection and normal RSS parsing. I’ll remove the unrelated HTTP/retry changes so the security-relevant diff is easy to audit.

I’ll push the cleanup shortly and would appreciate another look once it’s updated.

Yijia-Xiao:
Thanks @anupamme. Python's `xml.etree.ElementTree` does not resolve external entities or fetch external DTDs (an external-entity reference fails as undefined), so the XXE class does not apply at this call site, and the feed is public HTTPS. We did not add `defusedxml` for that reason. The real residual is parsing an unbounded body from an untrusted endpoint, which we addressed in `5a26ae1`: both the RSS and JSON reads are capped at 5 MiB and degrade through the existing failure paths. #1276 is a duplicate of this one.

anupamme:
Thanks for the explanation and sorry for the noise.

PR #1207, opened 2026-08-06
Add opt-in Schwab + AnySearch data vendors, vendor provenance, and cross-source basis disclosure

kevinkda:
## Summary

An **opt-in, zero-behavior-change** set of data-vendor and observability improvements. Defaults are unchanged (yfinance for prices/indicators, existing news vendors), so nothing is affected unless a user explicitly opts in.

This PR bundles five related pieces, all following the existing `route_to_vendor` / `VENDOR_METHODS` three-layer architecture:

1. **Schwab market-data vendor** (OHLCV + technical indicators, US equities/ETFs).
2. **AnySearch global-news vendor** (live-only supplement, look-ahead-safe).
3. **Vendor provenance + CLI visibility** (which vendor actually served; startup config summary).
4. **Cross-source price-basis disclosure** in reflection and the verified snapshot.
5. **Data-vendor selection** via `TRADINGAGENTS_DATA_VENDOR` env var and an interactive CLI prompt.

---

## 1. Schwab market-data vendor (opt-in)

Opt-in vendor for `core_stock_apis` (OHLCV) and `technical_indicators`. Lazily imported, disabled by default. Enable via:

```python
config["data_vendors"]["core_stock_apis"] = "schwab,yfinance"
config["data_vendors"]["technical_indicators"] = "schwab,yfinance"
```

- **US equities / ETFs only.** Symbols whitelisted to `^[A-Z]{1,5}$`; anything else falls back to yfinance.
- **Read-only.** No orders, no account access — OHLCV + indicators only.
- **Auth:** schwab-py's `client_from_token_file` (**refresh-only**, never a browser login; safe headless). `SCHWAB_APP_KEY` / `SCHWAB_APP_SECRET` + a `token.json` (`SCHWAB_TOKEN_PATH` / `SCHWAB_NATIVE_TOKEN_PATH`, or default state locations). The token can come from **any** schwab-py login; TradingAgents implements no login flow of its own.
- **Dependency:** new optional `schwab` extra pinned to `schwab-py>=1.5.1,<1.6`; the import is deferred so a clean install without the extra never fails.

## 2. AnySearch global-news vendor (opt-in, live-only)

New `anysearch` vendor for `get_global_news` backed by `api.anysearch.com`. AnySearch results carry **no reliable publish date** (verified against the live `/v1/search` response), so the vendor is strictly **look-ahead-safe**: it serves only live windows and raises for historical/backtest dates, letting the router fall back to yfinance (which date-filters). Never leaks future news into a backtest. Uses the core `requests` dependency (no new extra); `ANYSEARCH_API_KEY` is optional (anonymous works with lower limits). Not wired into any default.

## 3. Vendor provenance + CLI visibility

`route_to_vendor` now records which vendor actually served each method and which primaries it skipped, and logs the resolution at info level. The CLI prints:

- an **"Effective configuration"** panel at startup (LLM provider/models/backend, per-category data vendor, and — for Bedrock — the resolved region + auth mode: bearer token vs AWS profile), and
- a **"Data source"** panel after the run showing the serving vendor and any fallback reason.

This exposes silent degrades (e.g. an unconfigured Schwab primary skipped in favor of yfinance) that previously left it unclear which source and Bedrock region/auth were in use.

## 4. Cross-source price-basis disclosure

Realized returns for reflection/alpha stay on yfinance split/dividend-adjusted prices (the correct outcome basis), but when the agent priced its decision on an unadjusted vendor (e.g. Schwab raw prices), the stored reflection now carries an explicit basis caveat and the verified snapshot discloses its yfinance-adjusted basis. This prevents a raw-vs-adjusted artifact on split/dividend tickers from silently poisoning the memory log or being mistaken for a data error. Default (yfinance) behavior is unchanged.

## 5. Data-vendor selection (env / CLI)

`TRADINGAGENTS_DATA_VENDOR` env var + an interactive CLI "Data Vendor" step (skipped with a `✓ ... from environment` notice when the var is set). Applied to both `core_stock_apis` and `technical_indicators`. Fully opt-in; invalid vendor names fail loudly at startup.

---

## Testing

- `tests/test_schwab_vendor.py` (~35 tests + subtests, fully mocked — no real credentials).
- `tests/test_schwab_enum_contract.py` — pins the schwab-py `PriceHistory` enum wire-values the vendor sends (`PeriodType.YEAR`, `Period.FIVE_YEARS`, `FrequencyType.DAILY`, `Frequency.DAILY`); `importorskip` when schwab-py is absent.
- `tests/test_anysearch_vendor.py` — live-only guard, error classification, dedup, header/auth (mocked HTTP).
- `tests/test_price_basis_caveat.py` — basis caveat + snapshot disclosure.
- `tests/test_vendor_routing.py` — new provenance assertions (first-choice, fallback-with-reason, info log).
- `tests/test_data_vendor_env.py`, updated `tests/test_cli_env_skip.py`, updated `tests/test_memory_log.py`.
- **New CI job** installs the `[schwab]` extra and runs the Schwab vendor + enum-contract tests, so schwab-py enum drift fails loudly instead of in production.
- Locally: full `pytest -q` → **648 passed, 1 skipped**; `ruff check .` → all checks passed. (Python 3.10–3.13 in CI.)

## Notes

- **Version intentionally not bumped** — release cadence is the maintainers' call. Changelog entries are under `## [Unreleased]` (Keep a Changelog format).
- **Candle timezone semantics not live-verified** for Schwab (epoch-ms day-boundary ET vs UTC); flagged as a `# TODO` and a known limitation. The frame is tz-naive and collapsed to a calendar day, so it only matters at cross-day boundaries.

Made with [Cursor](https://cursor.com)


Yijia-Xiao:
Thanks @kevinkda. The provenance half is addressed: each fundamentals vendor states the vintage rule it serves under, and SEC EDGAR serves US filings as filed (f881c4a). We are not adding the Schwab and AnySearch vendors: both need an account, and the default path stays keyless.

PR #1210, opened 2026-08-07
fix(researchers): guard bull/bear prompts against empty current_respo…

taro0915:
…nse (#1176)

`investment_debate_state["current_response"]` starts as `""`, and the graph fixes a speaking order, so whichever researcher runs first was handed a bare `Last bear argument:` / `Last bull argument:` with nothing after it. The model reads that as an argument it is supposed to rebut and invents one, opening the debate by arguing against a position the other side never took.

Build the opponent-argument line conditionally: when `current_response` is empty, state the absence explicitly instead of rendering an empty labelled field. The wording mirrors the guard the risk debators already carry ("If there are no responses from the other viewpoints yet, present your own argument based on the available data").

Applied to both researchers rather than only the bull. The bull is the opener under the current wiring, but the bear reaches the same state under a reversed debate order or from a state built without a prior bull turn, and the guard should not depend on graph wiring.

This removes the empty-field trigger; it does not claim to eliminate fabrication in general. #1176 notes the risk debators still fabricate despite carrying an equivalent instruction, so the prompt-level cue alone is not reliable. The remaining rebuttal cues in the opener are left untouched to keep this change minimal.

Adds tests/test_researcher_empty_response.py, which asserts against the rendered prompt each researcher actually sends: no empty labelled field on the opening turn, and a real opponent argument still passes through unchanged.

### DIFF
--- a/tests/test_researcher_empty_response.py
+++ b/tests/test_researcher_empty_response.py
@@ -1,0 +1,85 @@
+"""The debate opener must not be handed an empty opponent argument (#1176).
+
+`current_response` starts as `""`, and the graph fixes a speaking order, so
+whichever researcher runs first used to receive a bare ``Last bear argument:``
+(or ``Last bull argument:``) with nothing after it — which the model reads as
+an argument it should rebut, and fabricates one to rebut.
+
+These assert against the *rendered* prompt each researcher actually sends.
+"""
+from __future__ import annotations
+
+from unittest.mock import MagicMock
+
+import pytest
+
+from tradingagents.agents.researchers.bear_researcher import create_bear_researcher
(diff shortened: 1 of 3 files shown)


taro0915:
Noting for maintainers: #1200 covers the same issue and predates this PR. I should have linked it here when I opened this.

The two things this one adds are: the guard is applied to the bear researcher as well as the bull, since the bear reaches the same empty state under a reversed debate order or a state built without a prior bull turn; and tests/test_researcher_empty_response.py asserts against the rendered prompt each researcher actually sends, rather than the state dict.

If #1200 is the preferred base, I'm happy to close this and offer those two as a follow-up there instead. Whichever is less work to review.

taro0915:
Closing as #1200 covers this. Thanks.

PR #1236, opened 2026-08-15
feat(cli): persist interactive preferences safely

hamzabudeir:
## Summary

- Save CLI preferences so the same setup choices do not have to be entered every time.
- Add `config show`, `config set`, and `config reset` commands.
- Let users reuse saved preferences, run a one-off setup, or reconfigure them.
- Handle incomplete, invalid, or newer preference files safely.
- Preserve the existing Windows no-console behavior.
- Add tests and update the README and changelog.

## Motivation

I found it frustrating to sit through the same setup questions and select the same settings every time I ran the CLI. This change saves those choices so the setup only needs to be done once, while still allowing users to change them or run the setup again when needed.

## What it looks like
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -186,7 +186,10 @@ cython_debug/
 #   be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #   and can be added to the global gitignore or merged into this file.  For a more nuclear
 #   option (not recommended) you can uncomment the following to ignore the entire idea folder.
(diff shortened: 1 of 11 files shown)


Yijia-Xiao:
Thanks @hamzabudeir. Addressed differently: the CLI now offers your previous run's selections as prompt defaults, validated against the current catalog, rather than a stored preferences file with config commands.

PR #1250, opened 2026-08-23
fix(cli): integrate checkpoint resume

### LINKED ISSUE #1249: CLI --checkpoint flag does not actually enable checkpoint resume
The CLI exposes `--checkpoint` and `TRADINGAGENTS_CHECKPOINT_ENABLED`, but the interactive CLI execution path does not actually
configure or use the LangGraph checkpointer.

`TradingAgentsGraph.propagate()` configures the SQLite checkpointer, but `cli/main.py::run_analysis()` bypasses `propagate()` and
directly calls:

```python
graph.graph.stream(init_agent_state, **args)
 ```

 The graph is therefore compiled without a checkpointer and without a stable thread_id.

 ### Reproduction

 1. Run the CLI with checkpointing enabled:
    ```--checkpoint```
    or set:
    ```TRADINGAGENTS_CHECKPOINT_ENABLED=true```
 2. Start an analysis for a ticker and date.
 3. Interrupt or crash the run after at least one graph node completes.
 4. Run the same analysis again with the same ticker, date, analyst selection, and research depth.

 ### Actual behavior

 - No resumable checkpoint is created by the CLI path.
 - The second run starts from the beginning.
 - The CLI does not report that it is resuming from a checkpoint.
 - The checkpoint flag is accepted, but has no effect on the actual CLI graph execution.

 ### Expected behavior

 When checkpointing is enabled:

 - The CLI should compile the graph with the SQLite checkpointer.
 - It should provide a deterministic thread_id for the ticker, date, and graph shape.
 - A fresh run should receive the initial state.
 - A resumed run should pass None as the graph input so LangGraph resumes from the latest checkpoint.
 - Checkpoints should be cleared after successful completion.

 ### Root cause

 The checkpoint configuration exists, and the programmatic propagate() path sets up the checkpointer. However, the CLI has a
 separate streaming path for its Live UI and does not call the checkpoint setup logic before calling graph.stream().

 ### Regression test

 A CLI regression test should:

 1. Use a two-node graph.
 2. Crash in the second node on the first run.
 3. Assert that the first node ran once and a checkpoint exists.
 4. Run again with the crash disabled.
 5. Assert that the first node still ran only once, proving the second run resumed rather than replayed it.
 6. Assert that the checkpoint is cleared after successful completion.

Jaswanth-Sriram-Veturi:
Fixes #1249 

### Problem

The CLI accepted `--checkpoint`, but `run_analysis()` directly streamed a graph compiled without a checkpointer or stable
`thread_id`. Crashed CLI runs therefore could not resume.

### Changes

- Configure the SQLite checkpointer on the CLI streaming path.
- Propagate the deterministic checkpoint `thread_id`.
- Pass `None` when resuming so completed nodes are not replayed.
- Clear checkpoints after successful completion.
- Close the checkpointer in `finally`.
- Expose checkpoint lifecycle methods as public graph methods because the CLI uses them.
- Add a regression test that asserts completed nodes execute only once across crash and resume.

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -1127,131 +1127,157 @@ def run_analysis(checkpoint: bool | None = None):
         # (LLM tracking is handled separately via LLM constructor)
         args = graph.propagator.get_graph_args(callbacks=[stats_handler])
 
-        # Stream the analysis
-        trace = []
-        for chunk in graph.graph.stream(init_agent_state, **args):
-            # Process all messages in chunk, deduplicating by message ID
-            for message in chunk.get("messages", []):
-                msg_id = getattr(message, "id", None)
-                if msg_id is not None:
-                    if msg_id in message_buffer._processed_message_ids:
-                        continue
-                    message_buffer._processed_message_ids.add(msg_id)
-
-                msg_type, content = classify_message_type(message)
-                if content and content.strip():
-                    message_buffer.add_message(msg_type, content)
-
-                if hasattr(message, "tool_calls") and message.tool_calls:
-                    for tool_call in message.tool_calls:
-                        if isinstance(tool_call, dict):
-                            message_buffer.add_tool_call(tool_call["name"], tool_call["args"])
-                        else:
-                            message_buffer.add_tool_call(tool_call.name, tool_call.args)
-
-            # Update analyst statuses based on report state (runs on every chunk)
-            update_analyst_statuses(
-                message_buffer,
-                chunk,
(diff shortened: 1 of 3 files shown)


Yijia-Xiao:
Thanks @Jaswanth-Sriram-Veturi for the fix. Superseded by `51a245d`, which factors the checkpoint lifecycle into reusable methods shared by propagate and the CLI.

PR #1253, opened 2026-08-24
perf(graph): run the four analysts in parallel

### LINKED ISSUE #1255: Analysts run sequentially, making the analyst phase unnecessarily slow
## Summary

The four analysts (Market, Sentiment, News, Fundamentals) run as a single sequential chain, so the analyst phase takes as long as the **sum** of all four even though the analysts are independent of one another. This is the dominant latency component of a run.

## Current behavior

In `graph/setup.py`, analysts are wired back-to-back:

```
START → Market → (tools loop) → Msg Clear → Sentiment → … → Fundamentals → Bull Researcher
```

Each analyst only starts after the previous one's `Msg Clear` node fires. They share a single `messages` channel, cleared between each analyst — which is the only reason they cannot run concurrently.

## Why they can run in parallel

The analysts are independent:

- Each reads only the shared run inputs (ticker, date, resolved identity).
- Each writes exactly one `*_report` field and never reads another analyst's output.
- The `Msg Clear` step between them proves no cross-analyst message history is needed.

The single shared `messages` channel is used only as a per-analyst ReAct scratchpad, so the sole blocker to concurrency is message isolation.

## Proposed change

- Wrap each analyst in its own compiled ReAct subgraph with a **private `messages` channel**, so analysts cannot clobber each other's tool scratchpad.
- Fan out from `START` to all selected analysts in a single superstep, and fan back in at the Bull Researcher (which runs once after all reports are in).
- Only each analyst's `*_report` field crosses back to the parent graph.
- Remove the now-unnecessary `Msg Clear` nodes (the subgraph `END` discards messages).

Downstream nodes (researchers, trader, risk team, portfolio manager) are unchanged.

## Impact

Analyst wall-time drops from **sum → max** of the four analysts, with no change to outputs.

## Scope / non-goals

- Intra-run analyst parallelism only. Cross-run / cross-ticker concurrency is out of scope.
- Behavior of all downstream stages is unchanged.

## Notes / trade-offs

- **Checkpoint resume:** a subgraph invocation is atomic to the parent checkpointer, so on a `--checkpoint` resume an unfinished analyst re-runs from scratch rather than mid-loop. Acceptable — analysts simply re-fetch.
- **CLI wall-time tracker** assumes sequential execution; under parallelism its per-analyst timings become approximate (not broken).

## Testing

- New unit tests covering message isolation and the fan-in barrier (no LLM required).
- Full existing suite passes with no regressions.

## References

- Implemented by #1253

hailampy123:
## What

The market, sentiment, news, and fundamentals analysts are independent — each reads only the shared run inputs (ticker, date, resolved identity) and writes exactly one `*_report` field, never another analyst's output. They previously ran as a **single sequential chain** sharing one `messages` channel (cleared between each), so end-to-end analyst latency was the **sum** of all four.

This wraps each analyst in its own compiled ReAct subgraph with a **private `messages` channel** (`graph/analyst_subgraph.py`). The parent graph now **fans out** from `START` to every selected analyst in one superstep and **fans in** at the Bull Researcher.

## Why it's safe

- Only each analyst's `report_key` crosses back to the parent — the tool-call scratchpad stays inside the subgraph and is discarded — so concurrent analysts **cannot clobber** each other's messages.
- The `Msg Clear` nodes are removed (subgraph `END` discards messages instead).
- Downstream nodes (researchers, trader, risk team, portfolio manager) are **unchanged**.

## Impact

Analyst wall-time drops from **sum → max** of the four analysts — the largest latency component of a run — for a roughly 2–4× speedup of the analyst phase, no change to outputs.

## Caveats

- **Checkpoint resume:** `subgraph.invoke()` is atomic to the parent checkpointer, so on a `--checkpoint` resume an unfinished analyst re-runs from scratch rather than mid-loop. Acceptable — analysts just re-fetch.
- **CLI wall-time tracker** assumes sequential execution; under parallelism its per-analyst timings become approximate (not broken).

## Tests

- New `tests/test_analyst_parallel.py` (5 tests): message isolation + fan-in barrier, no LLM required.
- Full suite: **579 passed, 2 skipped** (skips unrelated: bedrock import, deepseek live key), no regressions.
- Verified the real graph compiles; edges confirmed: `START` → 4 analysts, 4 analysts → Bull Researcher.

Closes #1255

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,15 +103,11 @@ git clone https://github.com/TauricResearch/TradingAgents.git
 cd TradingAgents
 ```
 
-Create a virtual environment in any of your favorite environment managers:
+Create a virtual environment and install the package with [uv](https://docs.astral.sh/uv/):
 ```bash
-conda create -n tradingagents python=3.12
-conda activate tradingagents
-```
-
-Install the package and its dependencies:
-```bash
-pip install .
+uv venv --python 3.12
+source .venv/bin/activate  # Windows: .venv\Scripts\activate
+uv pip install .
 ```
 
 ### Docker
--- a/tests/test_analyst_parallel.py
+++ b/tests/test_analyst_parallel.py
@@ -1,0 +1,146 @@
+"""Analyst parallelization: fan-out/fan-in with isolated message channels.
+
+Verifies the core guarantee of the parallel-analyst refactor: concurrent
+analysts each get a fresh, private ``messages`` scratchpad (so they can never
+clobber each other), and every selected analyst's report reaches the parent
+state exactly once after a single fan-in barrier.
+
+No LLM is used — analyst nodes are stubbed. The stub emits no tool calls, so
+``should_continue`` immediately routes to the clear node (END inside the
+subgraph), exercising the real subgraph + wrapper wiring.
+"""
+
+import unittest
+
+from langgraph.graph import END, START, StateGraph
+
+from tradingagents.agents.utils.agent_states import AgentState
+from tradingagents.graph.analyst_execution import build_analyst_execution_plan
+from tradingagents.graph.analyst_subgraph import (
+    build_analyst_subgraph,
+    make_analyst_wrapper,
+)
+
+
+def _stub_analyst(report_key, marker):
+    """A no-LLM analyst node: records what messages it saw, writes its report."""
+
+    def node(state):
+        seen = " | ".join(str(getattr(m, "content", m)) for m in state["messages"])
+        # Emit a plain AI message with no tool calls -> should_continue -> END.
+        return {
+            "messages": [("ai", f"{marker} done")],
+            report_key: f"{marker}::saw[{seen}]",
+        }
+
+    return node
+
+
+def _stub_should_continue(spec):
+    """Mirror ConditionalLogic.should_continue_*: tools if tool_calls, else clear."""
+
+    def should_continue(state):
+        last = state["messages"][-1]
+        if getattr(last, "tool_calls", None):
+            return spec.tool_node
+        return spec.clear_node
+
+    return should_continue
+
+
+class AnalystWrapperIsolationTests(unittest.TestCase):
+    def test_wrapper_returns_only_report_and_seeds_fresh_messages(self):
+        spec = build_analyst_execution_plan(["market"]).specs[0]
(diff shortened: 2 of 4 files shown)


hamzabudeir:
Hey, that's pretty dope!

vipxzx:
> ## What
> 
> The market, sentiment, news, and fundamentals analysts are independent — each reads only the shared run inputs (ticker, date, resolved identity) and writes exactly one `*_report` field, never another analyst's output. They previously ran as a **single sequential chain** sharing one `messages` channel (cleared between each), so end-to-end analyst latency was the **sum** of all four.
> 
> This wraps each analyst in its own compiled ReAct subgraph with a **private `messages` channel** (`graph/analyst_subgraph.py`). The parent graph now **fans out** from `START` to every selected analyst in one superstep and **fans in** at the Bull Researcher.
> 
> ## Why it's safe
> 
> - Only each analyst's `report_key` crosses back to the parent — the tool-call scratchpad stays inside the subgraph and is discarded — so concurrent analysts **cannot clobber** each other's messages.
> - The `Msg Clear` nodes are removed (subgraph `END` discards messages instead).
> - Downstream nodes (researchers, trader, risk team, portfolio manager) are **unchanged**.
> 
> ## Impact
> 
> Analyst wall-time drops from **sum → max** of the four analysts — the largest latency component of a run — for a roughly 2–4× speedup of the analyst phase, no change to outputs.
> 
> ## Caveats
> 
> - **Checkpoint resume:** `subgraph.invoke()` is atomic to the parent checkpointer, so on a `--checkpoint` resume an unfinished analyst re-runs from scratch rather than mid-loop. Acceptable — analysts just re-fetch.
> - **CLI wall-time tracker** assumes sequential execution; under parallelism its per-analyst timings become approximate (not broken).
> 
> ## Tests
> 
> - New `tests/test_analyst_parallel.py` (5 tests): message isolation + fan-in barrier, no LLM required.
> - Full suite: **579 passed, 2 skipped** (skips unrelated: bedrock import, deepseek live key), no regressions.
> - Verified the real graph compiles; edges confirmed: `START` → 4 analysts, 4 analysts → Bull Researcher.

vipxzx:
Sex

Issue #1251, opened 2026-08-23
Decision-log lessons lack a point-in-time guard — possible lookahead in backtests


fabiolenine:
# Decision-log lessons lack a point-in-time guard — possible lookahead in backtests

**Type:** enhancement (possible correctness bug) · **Area:** memory / reflection

## Summary

Decision-log reflections are grounded on *realized* outcomes and later re-injected into agent prompts, but recall is keyed by ticker and recency with no constraint that a lesson's outcome pre-dates the current analysis date. When runs are not executed in strict chronological order (backtesting a date after already running a later one, or re-running), a run for date `D1` can be conditioned on a lesson derived from an outcome resolved on `D2 > D1`. That's lookahead, and it can inflate apparent alpha.

## Where

- `tradingagents/agents/utils/memory.py` → `TradingMemoryLog.get_past_context()` selects entries by reverse append order, filtering only by `pending` status and ticker — no comparison against the current run's `trade_date`.
- `tradingagents/graph/trading_graph.py` → `_run_graph()` injects `get_past_context(company_name)` as `past_context`.
- `tradingagents/agents/managers/portfolio_manager.py` → consumes `past_context` as "Lessons from prior decisions and outcomes" in the PM prompt.

Because the store is ordered by *run* time, not by the *trade_date* being analyzed, resolved lessons from chronologically-later trade dates are eligible for injection into an earlier-dated run.

## Why it matters

This compounds the documented behavior that news/social sources "reflect now" for a historical analysis date. Together they mean a backtest run can be influenced by information from its own future, undermining the validity and reproducibility of any backtested return figure.

## Minimal fix (works with the current markdown store)

1. Ensure each resolved entry carries the **outcome-resolution date** (the tag already gains return figures on resolution; add/keep the resolution date).
2. In `get_past_context`, accept the current `trade_date` and filter to entries whose outcome-resolution date is **strictly before** it. Pending entries are already excluded, so this is a small, local change.

That alone closes the leak without altering the `pending → resolved` lifecycle. I'm on `main` (v0.3.0) and glad to submit a PR for this if the approach looks right.

---

## Broader proposal (optional, follow-up)

[...]

PR #1254, opened 2026-08-25
fix(memory): keep future-resolved lessons out of point-in-time prompts

hudsonwa:
Hi! First time contributing here, so thanks in advance for your patience with me.

Thank you as well to whoever wrote up #1251 so carefully; the file pointers made it genuinely easy to dig into, and out-of-order backtests are such an easy trap to fall into when you're iterating on dates.

I've had a go at a small fix. The idea: `get_past_context()` now takes an optional `as_of_date`, and any lesson whose outcome resolved after that date simply doesn't get shown to the agents. Same-day lessons stay in, since those are fair game. `trading_graph.py` passes the run's trade date through automatically, so nobody has to remember to opt in. And if you don't pass a date at all, nothing changes, existing behaviour stays exactly as it was.

Entries with unparseable dates are excluded rather than trusted, which seemed the safer side to land on, but happy to flip that if you'd prefer it to fail loud instead.

Tests: five new ones covering future exclusion, same-day inclusion, cross-ticker lessons, backwards compatibility, and malformed dates. Confirmed they fail against main without the change and pass with it, and the full suite is green (581 passed, 2 pre-existing skips).

Really enjoy what this project is building. Happy to tweak whatever doesn't fit the direction you want here. Cheers!

### DIFF
--- a/tests/test_memory_log.py
+++ b/tests/test_memory_log.py
@@ -284,6 +284,52 @@ class TestTradingMemoryLogCore:
         assert "AAPL" not in ctx
         assert "META" in ctx
 
+    # Point-in-time guard (as_of_date)
+
+    def test_as_of_date_excludes_future_resolved_entries(self, tmp_path):
+        """An entry resolved after the current trade date must not leak in.
+
+        This is the lookahead guard: running date D1 after a later run for
+        D2 would otherwise inject D2's reflection (which encodes knowledge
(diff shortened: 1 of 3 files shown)


Yijia-Xiao:
Thanks @hudsonwa. Superseded by `8db41f6` (point-in-time filter on the stored resolution date).

PR #1266, opened 2026-08-27
Add point-in-time data provenance and research contracts

ZahirBodrike:


Yijia-Xiao:
Thanks @ZahirBodrike. Point-in-time provenance is now in the data layer: SEC EDGAR serves US statements as filed (f881c4a), vendors that cannot date a figure by its filing say so in the output they return, and a window a feed never observed is reported as unavailable rather than as an absence.

PR #1277, opened 2026-08-30
fix(dataflows): pin FRED requests to curr_date vintage to remove look-ahead bias

### LINKED ISSUE #1275: [Bug] FRED macro vendor returns unpublished and revised data in historical backtests (silent look-ahead bias)
## Summary

`get_macro_data` bounds its FRED requests by observation date only. The vintage
window is a second, independent pair of parameters, and it defaults to today. So
a run dated 2023-03-31 receives March CPI, March payrolls, March unemployment
and Q1 GDP, none of which had been published then, at values that have since
been revised. Nothing errors and the output looks normal.

Different vendor from the yfinance news leak in #1007, fixed by `0c1231a`. I
found it while reading #805.

`tradingagents/dataflows/fred.py` → `get_macro_data()`, reached by the agent as
the `get_macro_indicators` tool. Present on `main` at `a33fd4c`; the file is 237
lines and the repository has no `realtime` references anywhere.

## Steps to reproduce

```python
from tradingagents.dataflows.fred import get_macro_data
print(get_macro_data("cpi", "2023-03-31", look_back_days=90))
```

Returns the observation dated 2023-03-01. FRED released it on 2023-04-12. Same
for payrolls, unemployment, industrial production and real GDP.

## Cause

`tradingagents/dataflows/fred.py:180-188`:

```python
observations = _request(
    "series/observations",
    {
        "series_id": series_id,
        "observation_start": start_date,
        "observation_end": curr_date,
        "sort_order": "asc",
    },
).get("observations", [])
```

`series/observations` takes two independent windows. `observation_start` /
`observation_end` bound the period the data describes. `realtime_start` /
`realtime_end` bound the vintage, and are documented as *"optional, default:
today's date"*. Only the first pair is set, so every historical request is
served from today's vintage. A series describing March is released in April, so
an observation-date bound admits it.

The docstring at `tradingagents/dataflows/fred.py:146-147` intends the opposite:

```
curr_date: End of the window (yyyy-mm-dd); no later observations are
    returned, so a past date never leaks future data.
```

True of observations. Not true of information.

`curr_date` is the run's own trade date.
`tradingagents/agents/analysts/news_analyst.py:15` binds
`current_date = state["trade_date"]`, the system prompt tells the model to treat
it as "now" when setting tool date ranges, and the resulting
`get_macro_indicators` call reaches the request above via
`tradingagents/dataflows/interface.py:138` (`"fred": get_fred_macro_data`,
aliased to `get_macro_data` at `interface.py:20`).

## Impact

Measured against the live API for a decision date of 2023-03-31 over a one year
window, fetching each series twice: once with the vendor's request parameters,
once with the vintage pinned to the decision date.

Every series returns exactly one release that did not exist yet:

| Series | Returned | Published by 2023-03-31 | Leaked observation |
|---|---|---|---|
| GDPC1 | 5 | 4 | Q1 2023 real GDP |
| PAYEMS | 13 | 12 | March payrolls |
| CPIAUCSL | 13 | 12 | March CPI |
| UNRATE | 13 | 12 | March unemployment |
| INDPRO | 13 | 12 | March industrial production |

Observations present in both vintages have moved as well, which is revision
effect on its own:

```
GDPC1   2022-01-01   known then:  19924.088   returned now:  21932.71
PAYEMS  2023-01-01   known then:     155039   returned now:     154776
INDPRO  2022-12-01   known then:   101.6086   returned now:    99.7664
UNRATE  2023-01-01   known then:        3.4   returned now:        3.5
```

Q1 2022 real GDP reads about 10% higher today, largely rebasing. January 2023
payrolls are down 263,000.

What I have not measured is what any of this is worth to a reported result.
Future information reaching the decision context and that information changing
the outcome are two different claims, and I have only established the first.
These series feed qualitative commentary rather than a numeric signal, so the
effect could be anywhere from negligible to substantial. I would rather not
guess.

## Fix

The obvious one-line version is wrong in two ways, so the patch below is what I
tested rather than what I first wrote.

`fred.py:168`, the `series` metadata call, needs pinning too. That endpoint is
vintage-sensitive: GDPC1's units read `Billions of Chained 2012 Dollars` in the
2023-03-31 vintage and `2017 Dollars` today. Pinning only the observations
returns correct 2012-dollar values under a 2017-dollar header, which is worse
than the current behaviour because it looks consistent.

The pin also has to be conditional. FRED rejects a `realtime_start` after its
own today (US Central) with a 400, which `_request` surfaces as a `ValueError`,
and `curr_date` arrives from the caller's *local* clock (`cli/main.py:566`
defaults to `datetime.datetime.now()` and validates against local `now()`). An
unconditional pin crashes live runs for every user ahead of US Central.
Unpinned is already correct there: it means latest published.

```python
from datetime import date, datetime, timedelta   # `date` is new

    # Pin the vintage so the response carries only what FRED had published by
    # the decision date, at the values it published then. Both requests need
    # it: the `series` metadata endpoint is vintage-sensitive too.
    #
    # Only pin a past date. FRED rejects a realtime date after its own today
    # (US Central) with a 400, and curr_date comes from the caller's LOCAL
    # clock, so an unconditional pin would crash a live run for a user ahead
    # of US Central. Unpinned is already right for live use: latest vintage.
    pit = (
        {"realtime_start": curr_date, "realtime_end": curr_date}
        if curr_date < date.today().isoformat()
        else {}
    )

    meta = _request("series", {"series_id": series_id, **pit}).get("seriess") or []

    # ... and in the observations request:
    observations = _request(
        "series/observations",
        {
            "series_id": series_id,
            "observation_start": start_date,
            "observation_end": curr_date,
            "sort_order": "asc",
            **pit,
        },
    ).get("observations", [])
```

Applied to a clean clone of `a33fd4c`:

- `get_macro_data("cpi", "2023-03-31", 90)` no longer returns the 2023-03-01
  observation; latest becomes 301.648 (2023-02-01).
- `get_macro_data("real_gdp", "2023-03-31")` returns 2012-dollar values under a
  `Bil. of Chn. 2012 $` header, so the two agree.
- `curr_date` set to a local tomorrow still succeeds instead of raising.

One thing I have not resolved for you: ALFRED vintage coverage starts at a
different date for each series and a few series have none, so a pin can return
an empty set where the current call returned data. The empty-result branch at
`fred.py:205-209` probably needs to tell "no vintage available" apart from
"window too short", or the change will present as a data outage.

Happy to open a PR with this plus regression tests if that is useful. The
reproduction is a standalone script that queries each series both ways, no model
calls, and runs in seconds on a free API key.

---

I went looking here because I shipped this same class of defect into my own
production system, reading macro series by observation date when only the prior
day's publication was actually available. It cost me a programme of results and
a retraction, so it is the first place I check now.

Jackzigen:
## Summary

Fixes #1275.

`get_macro_data()` queried FRED without realtime bounds, so historical backtests silently ingested as-of-today values: unreleased prints (e.g. CPI flash vs final) and later revisions — classic silent look-ahead bias.

This PR:

- Pins both the `series` and `series/observations` requests to the `curr_date` ALFRED vintage via `realtime_start` / `realtime_end`, so a backtest only sees values that were actually published by that date
- Dedups multiple vintages per observation date (revised series otherwise render one row per vintage); the latest vintage in the window wins
- Clarifies the empty-result message: under realtime bounds an empty window usually means no ALFRED vintage coverage for that series, not just a short window

## Testing

- `tests/test_fred.py`: 18/18 passing, including 4 new regression tests (vintage-pinned request params, revision dedup keeps latest vintage, empty-window message, existing lookahead-safe window)
- All tests fully mocked — no FRED key or network needed in CI
- Full test suite shows no new failures (remaining local failures are pre-existing dependency-mismatch issues unrelated to this change, verified by stashing the diff)

## Notes

Vintage depth varies per series on ALFRED; series without vintage coverage now return an explicit "no vintage data" message instead of silently returning current revised values.

### DIFF
--- a/tests/test_fred.py
+++ b/tests/test_fred.py
@@ -161,11 +161,33 @@ class FredFormattingTests(unittest.TestCase):
             captured[path] = params
             return _META if path == "series" else _OBS
 
-        with mock.patch.object(fred, "_request", side_effect=_capture):
+        # Pin the clock so the clamp (min(curr_date, FRED's today)) is a no-op
+        # here: FRED's today == curr_date.
+        with mock.patch.object(fred, "_request", side_effect=_capture), \
+                mock.patch.object(fred, "_fred_today", return_value="2025-09-30"):
             fred.get_macro_data("unemployment", "2025-09-30", 90)
         for path in ("series", "series/observations"):
             self.assertEqual(captured[path]["realtime_end"], "2025-09-30", path)
             self.assertEqual(captured[path]["realtime_start"], "2025-07-02", path)
+
+    def test_realtime_end_is_clamped_to_fred_today(self):
+        # A curr_date ahead of FRED's clock (caller in Asia during their
+        # morning) must clamp realtime_end to FRED's today, not 400: FRED
+        # validates realtime dates against US Central time, and the routing
+        # layer turns that 400 into a silent DATA_UNAVAILABLE sentinel
+        # (#1275) - macro data would vanish without any loud failure.
+        captured = {}
+
+        def _capture(path, params):
+            captured[path] = params
+            return _META if path == "series" else _OBS
+
+        # Local 2025-10-31 in Asia == 2025-10-30 in US Central.
+        with mock.patch.object(fred, "_request", side_effect=_capture), \
+                mock.patch.object(fred, "_fred_today", return_value="2025-10-30"):
+            fred.get_macro_data("unemployment", "2025-10-31", 90)
+        self.assertEqual(captured["series/observations"]["realtime_end"], "2025-10-30")
+        self.assertEqual(captured["series"]["realtime_end"], "2025-10-30")
 
     def test_revised_observations_keep_latest_vintage(self):
         # With realtime bounds FRED returns one row per vintage; a revision
--- a/tests/test_fred.py
+++ b/tests/test_fred.py
@@ -121,10 +121,11 @@ class FredFormattingTests(unittest.TestCase):
         self.assertIn("not found", out)
 
     def test_long_series_is_truncated_but_change_uses_full_range(self):
-        # Build > MAX_ROWS observations deterministically.
+        # Build > MAX_ROWS observations deterministically (unique dates: the
+        # report deduplicates revisions by observation date).
         obs = {
             "observations": [
-                {"date": f"2025-01-{(i % 28) + 1:02d}", "value": str(i)}
+                {"date": f"2025-{(i // 28) + 1:02d}-{(i % 28) + 1:02d}", "value": str(i)}
                 for i in range(fred.MAX_ROWS + 10)
             ]
         }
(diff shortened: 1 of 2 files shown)


PyriteResearch:
Thanks for turning this round so fast, and for keeping the tests key-free.

The pin is unconditional. `vintage` at `fred.py:167-170` applies whatever
`curr_date` is. FRED rejects a realtime date after its own today (US Central)
with a 400, `_request` raises that as a `ValueError` at `fred.py:131`, and
`curr_date` comes from the caller's local clock (`cli/main.py:566`). Anyone
whose local date is ahead of FRED's hits it, which is most of the day in New
Zealand, Australia and much of Asia.

I overstated what that costs in #1275 and have corrected it there. It does not
crash: `macro_data` is in `OPTIONAL_CATEGORIES` (`interface.py:92`), so
`route_to_vendor` catches the error at `interface.py:214` and returns
`DATA_UNAVAILABLE` from `interface.py:253-258`. Those users just lose macro
data, silently, on every live run. The mocked tests cannot see it.

Guarding on a past date is enough. Unpinned is already right for a live run: it
means latest published.

```python
vintage = (
    {"realtime_start": start_date, "realtime_end": curr_date}
    if curr_date < date.today().isoformat()
    else {}
)
```

Smaller one. The docstring says values come back "at their as-published values"
and the empty branch says "no as-first-published vintage data", but the dedup
keeps the latest vintage per date. What you return is the value as known on
`curr_date`, not as first published, and those differ for anything revised in
between. The code is right and the two strings are wrong.

Optional: `realtime_start: start_date` is what forces the dedup, since it
returns a row per vintage per observation. `realtime_start: curr_date` gets the
same values one row per observation. Yours works, it just does more.

One caveat on all of it. The 400 and local-clock behaviour I tested on a clean
clone of `a33fd4c` when I wrote #1275. The routing trace I read from `main` and
from this PR at `963867c`, not executed. No FRED key on this machine today, so
nothing here was re-run live.

Jackzigen:
Thanks for the correction - and for tracing the routing path, that's the detail that matters here.

You're right about the failure mode, and I've confirmed it in the code: `_request` raises on FRED's 400 (`fred.py`), `route_to_vendor` catches it, and since `macro_data` sits in `OPTIONAL_CATEGORIES`, the call degrades to the `DATA_UNAVAILABLE` sentinel. Users ahead of US Central would lose macro data silently every morning.

Pushed a follow-up commit that makes the pin conditional as you described:

- `realtime_end = min(curr_date, FRED's today)` where FRED's today is computed in `America/Chicago`
- For past dates the pin is unchanged - clamping only ever applies to a date already in FRED's future, where no as-of-date data could exist anyway, so the look-ahead protection is intact
- New regression test pins `_fred_today()` and asserts the clamp; existing vintage tests pin the clock so the clamp is a no-op there

`tests/test_fred.py`: 19/19 passing, plus `test_vendor_routing.py` / `test_vendor_errors.py` (33/33 combined) to make sure the routing layer paths you pointed at are still green.

Yijia-Xiao:
Thanks @Jackzigen. Superseded by `8b7ece8`, which pins the FRED vintage on both the observations and metadata requests.

Issue #1286, opened 2026-09-01
Feature Request: Coordinate Reddit RSS rate-limit handling across threads/processes


Recnelis0:
**Problem**

The current implementation's 429 handling only protects a single sequential caller. Concurrency exposes two gaps:

1. **Same-process races (multiple threads).** Nothing serializes access to the rate-limit state. If `fetch_reddit_posts` is called from multiple threads (e.g. several ticker analyses running in parallel), each call fetches independently and retries independently. There's no shared "how much budget is left" signal, so two threads can both decide it's safe to fire at the same moment and both get `429`.

2. **Cross-process blindness (multiple processes).** Rate-limit state, such as it exists, lives only in memory for the lifetime of a single process. Spawned subprocesses (e.g. one process per ticker, or a scheduled job re-invoked per run) have no visibility into requests another process just made against the same per-IP/per-feed budget. Each process starts "cold" and can immediately collide with work another process already did.

The net effect: under any concurrent usage pattern, `429`s become common rather than rare, and the module's single fixed-delay retry (`_fetch_subreddit_rss(..., _retry=True)`) isn't enough to absorb that — it only accounts for one caller's own request, not what siblings are doing.

fabiooliveir:
Hi @Recnelis0 and maintainers,

I'd like to work on this issue. As noted in `reddit.py` line 301, coordinating the per-IP request budget and backoff state is essential when running multiple tickers in parallel or across separate worker processes.

Here is the proposed approach:
1. **In-process coordination (Threads):** Add a thread-safe rate limiter (`threading.Lock` + token/timestamp tracker) inside `reddit.py` so concurrent analysis threads pace requests cooperatively instead of racing into 429s.
2. **Cross-process coordination (Processes):** Use a lightweight file-backed state/lock in the cache directory (`~/.tradingagents/cache/reddit_pacing.json` or file lock) to record recent request timestamps and any active cooldown period across spawned processes.
3. **Shared cooldown on 429:** When a 429 occurs, record the `Retry-After` timestamp so subsequent calls across threads/processes immediately fast-degrade or pause instead of each process redundantly firing and stalling.
4. **Unit tests:** Comprehensive tests simulating multi-threaded calls and cross-process lock/cooldown behavior.

Let me know if this direction aligns with the framework's architecture, and I'll be glad to submit a PR!

Yijia-Xiao:
[...]

PR #1295, opened 2026-09-02
fix(reddit): report failed fetches as unavailable; raise headerless-429 back-off to 60s

sheiun-xu:
## Summary

Two related correctness fixes in `tradingagents/dataflows/reddit.py`, found while running v0.4.0 against NVDA and confirmed against the v0.4.1 base.

1. **A failed Reddit fetch was rendered as "no posts found".** `_fetch_subreddit_rss` returned `[]` for both "nothing matched" and "request failed", so a 429 produced the identical string as a genuinely empty search — and when every subreddit was throttled, the summary asserted there were no posts across all of them. The Sentiment Analyst treats that as a real signal: in a 2026-08-31 run it wrote *"r/stocks and r/investing are silent … keeps overall confidence low"* while the log showed both had returned `HTTP 429`. `stocktwits.py` already emits `<stocktwits unavailable: …>`; this brings Reddit in line.

2. **The headerless-429 back-off was an order of magnitude too short.** Measured against `/r/{sub}/search.rss`: a second request still 429s at 8 s, 10 s and 30 s of spacing and succeeds at 60 s. Reddit sends no `Retry-After`, so the 5 s fallback is what ran, which guaranteed the single retry also failed. Raising it to 60 s (jittered, headerless case only) recovers all three default subreddits in ~124 s where previously only the first request in a run succeeded. The `Retry-After` cap goes 30 s → 120 s for the same reason. `inter_request_delay` is intentionally unchanged: the back-off is adaptive and costs nothing when Reddit is not throttling.

Fix 1 is what made fix 2 diagnosable — once the failure stopped masquerading as an empty result, the retry that never succeeded became visible.

## Side effect worth knowing

`TestSentimentAnalystAgent` (5 tests, marked `unit`) only mocked the LLM and was fetching Yahoo, StockTwits and Reddit for real. With the longer back-off they went from ~15 s each to ~180 s each. Stubbing the three sources (the same seams `test_structured_agent_prompts.py` already stubs) drops the whole suite from ~100 s to ~5 s — most of the previous wall-clock was rate-limited network I/O.

## Evidence

| spacing between consecutive `search.rss` requests | 2nd request |
|---|---|
| 8 s / 10 s / 30 s | 429 |
| 60 s | 200 |

Before (2026-08-31 NVDA, r/stocks + r/investing both 429):
```
r/stocks: <no posts found mentioning NVDA in the past 7 days>
```
After, same conditions:
```
r/stocks: <unavailable: HTTP 429>
```
After, with the new back-off, default settings:
```
r/wallstreetbets — 5 recent posts …
r/stocks — 3 recent posts …
r/investing — 3 recent posts …
```

## Testing

- `pytest`: 660 passed, 2 skipped (bedrock / DeepSeek live) on top of v0.4.1 (`9dee508`)
- `ruff check`: clean
- Three pre-existing tests encoded the old silent-degradation contract by name (`…_fails_open`, `…_degrades_to_empty`, `…gives_up_after_one_retry`) and were updated; the v0.4.1 `test_oversized_rss_feed_is_refused_not_parsed` likewise now expects `RedditUnavailable`. Nine new tests cover the rate-limited / empty / all-failed / mixed renderings, the 60 s fallback band, and the widened `Retry-After` cap.

### DIFF
--- a/tests/test_reddit_fallback.py
+++ b/tests/test_reddit_fallback.py
@@ -86,9 +86,10 @@ class TestRssParsing:
         assert posts[0]["created_utc"] > 0
         assert "datacenter unit" in posts[0]["selftext"]
 
-    def test_malformed_xml_fails_open(self):
-        with patch.object(reddit, "urlopen", return_value=_resp(lambda: b"<<not xml>>")):
-            assert reddit._fetch_subreddit_rss("NVDA", "stocks", 5, 5.0) == []
+    def test_malformed_xml_raises_unavailable(self):
+        with patch.object(reddit, "urlopen", return_value=_resp(lambda: b"<<not xml>>")), \
+             pytest.raises(reddit.RedditUnavailable):
+            reddit._fetch_subreddit_rss("NVDA", "stocks", 5, 5.0)
 
 
 @pytest.mark.unit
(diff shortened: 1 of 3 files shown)


Yijia-Xiao:
Thanks @sheiun-xu — the measurements are what made this actionable, and both points were right.

Shipped in `7cc478a`. A failed fetch now returns a distinct result from an empty search and renders as unavailable, and the run summary only claims silence for subreddits actually searched; when every source fails it says so instead of asserting there were no posts. The back-off is 60s for the headerless case, as your spacing measurements showed a retry only starts succeeding there.

One thing we added on top: the back-off is paid at most once per run. Once one source has been throttled the per-IP budget is generally gone, so spending a further 60s on each remaining subreddit stalls the analysis without getting through. Cross-process coordination is tracked separately in #1286.

PR #1298, opened 2026-09-03
Fall back to Alpha Vantage when Yahoo Finance fails

jlx357890-cloud:
## What changed

- `tradingagents/default_config.py`: the `core_stock_apis`, `technical_indicators`, `fundamental_data`, and `news_data` vendor entries now use the ordered fallback chain `"yfinance,alpha_vantage"` instead of `"yfinance"` alone.
- `.env.example`: documents `ALPHA_VANTAGE_API_KEY`, which was previously undocumented even though it's read directly by `alpha_vantage_common.get_api_key()`.

## Why

Yahoo Finance has been intermittently unavailable/rate-limiting. Rather than switching the default vendor outright (Yahoo is free/keyless with generally looser limits), this chains Alpha Vantage as an automatic fallback so requests only go to Alpha Vantage when Yahoo fails, per the existing "ordered fallback" mechanism already documented in `default_config.py`.

## Notes for reviewers

- Alpha Vantage's free tier is fairly tight (25 requests/day, 5/min), so it's meant as a fallback, not a primary vendor.
- No behavior changes for users who don't set `ALPHA_VANTAGE_API_KEY` — `get_api_key()` already raises a clear error if a code path needs it and it's unset, and the fallback only engages after Yahoo fails.
- This PR does not include unrelated local changes present in the working tree (a `stocktwits.py` WAF-bypass fix and a local `.env`-derived scratch file); those are intentionally out of scope here.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -19,6 +19,10 @@ NVIDIA_API_KEY=
 # FRED (Federal Reserve macro data: rates, inflation, labor, growth). Free key: https://fred.stlouisfed.org/docs/api/api_key.html
 #FRED_API_KEY=
 
+# Alpha Vantage (stock/fundamentals/news fallback when Yahoo Finance is
+# unavailable). Free key: https://www.alphavantage.co/support/#api-key
+#ALPHA_VANTAGE_API_KEY=
+
 # Optional: a custom OpenAI-compatible endpoint (vLLM, LM Studio, llama.cpp,
 # relay). Select provider "openai_compatible" and set the base URL; the key is
 # optional (local servers need none).
--- a/tradingagents/default_config.py
+++ b/tradingagents/default_config.py
@@ -137,10 +137,10 @@ DEFAULT_CONFIG = _apply_env_overrides({
     # routed to vendors you didn't choose. For ordered fallback, list several,
     # e.g. "yfinance,alpha_vantage". "default" uses all available vendors.
     "data_vendors": {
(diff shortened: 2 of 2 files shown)


fabiooliveir:
Hi @jlx357890-cloud and maintainers,

Thanks for opening this PR! Chaining Alpha Vantage as an ordered fallback (`"yfinance,alpha_vantage"`) is a great improvement for handling Yahoo Finance rate limits and intermittent outages.

While analyzing the fallback mechanism, we identified an underlying issue that currently prevents the fallback from triggering for fundamentals and news:

In `tradingagents/dataflows/y_finance.py` (`get_fundamentals`, `get_balance_sheet`, `get_cashflow`, `get_income_statement`) and `yfinance_news.py` (`get_news_yfinance`, `get_global_news_yfinance`), exceptions are caught internally and returned as plain error strings:
```python
except Exception as e:
    return f"Error retrieving fundamentals for {ticker}: {str(e)}"
```
Because these functions return a string instead of letting the exception propagate, `interface.route_to_vendor` treats the return as a successful data response and never engages the fallback to Alpha Vantage.

To make the fallback chain truly operational across all tools, these functions need to propagate the underlying exception so `route_to_vendor` can catch it and proceed down the vendor chain. We also need regression unit tests covering the fallback transition when Yahoo fails.

We are preparing a branch with this fix and the associated unit tests. Would you prefer a PR into your branch, or should we submit a companion PR referencing #1298?

Issue #1166, opened 2026-07-25
[Feature Request] Add a first-class portfolio context input to AgentState and graph entrypoints


lx7720:
## Problem

TradingAgents provides Trader, risk-management, and Portfolio Manager nodes, but portfolio state is not a first-class input in the public graph contract. The current AgentState and canonical graph entrypoint do not define a documented, typed way to pass current holdings, allocation constraints, available capital, or existing-position status to the agents.

This makes downstream integrations error-prone: a wrapper can persist a portfolio snapshot in an output artifact while the graph state consumed by Trader and Portfolio Manager remains empty. The run still completes and produces position-sizing language, but those recommendations are not necessarily grounded in the actual portfolio.

In practice, when a wrapper fails to populate this context, the Trader or Portfolio Manager can still produce percentage-based sizing language (for example, "+1-2% overweight") without indicating that no position data was available. The output can therefore look like a portfolio-grounded recommendation even when the graph had no portfolio context.

## Why this matters

- "Maintain", "trim", "add", and "overweight" decisions require current-position context.
- Missing context should not be indistinguishable from an empty portfolio.
- Text-only or wrapper-specific injection can make CLI and Python API behavior diverge.
- Checkpoints and saved reports should identify whether portfolio context was present and which snapshot was used.

## Possible direction

One possible direction would be to introduce an optional, typed portfolio-context contract at the canonical graph entrypoint and make it available to the nodes that make or validate portfolio decisions. The maintainers may prefer a different field layout or propagation mechanism; the key requirement is that the public contract make the presence or absence of portfolio context explicit.

Depending on the chosen design, the context might cover:

- instrument identity and current position status
- portfolio and allocation constraints
- available capital or sizing limits
- snapshot timestamp and source
- an explicit "missing / not provided" state

For position-sizing decisions, missing context could produce a visible warning or a safe "context unavailable" result rather than silently generating percentage recommendations.

## Acceptance criteria

- A documented portfolio context can be passed through both Python and CLI entrypoints.
- The same context reaches all nodes that make or validate transaction or portfolio decisions.
- Saved state and report metadata record whether context was present and identify the snapshot used.
- Tests cover both a populated context and an explicitly missing context.
- No broker connection or order execution is required; this is an input-contract and validation issue.

This request is intentionally based on the public API contract and does not attach account data or private analysis reports.

ihsieh31:
I have a small broker-neutral implementation of this direction working locally against current main.

[...]

PR #1304, opened 2026-09-05
feat: add first-class portfolio context to graph inputs

ihsieh31:
## Summary

Adds an optional, broker-neutral portfolio context to the canonical TradingAgents graph input so Trader / risk / Portfolio Manager decisions can distinguish between:

- a known current position / portfolio snapshot
- a known empty portfolio
- portfolio context that was not provided

This addresses the input-contract gap discussed in #1166.

## Why

TradingAgents can currently produce sizing and Buy/Overweight/Hold/Underweight/Sell guidance, but the graph does not have a first-class typed representation of current holdings and capital.

Downstream wrappers can inject this information through prose, but that makes behavior inconsistent and can make a missing portfolio look like a flat portfolio.

## Scope

This PR intentionally does **not** add:

- broker connectivity
- order execution
- Alpaca dependencies
- automatic trading
- portfolio optimization

It only adds the portfolio input contract and threads it through the existing decision flow:

- `PositionSnapshot` / `PortfolioContext` schemas plus a deterministic renderer focused on the instrument under analysis (`tradingagents/agents/schemas.py`)
- `AgentState.portfolio_context` carried as JSON-safe data so checkpoints stay serializable
- `propagate(..., portfolio_context=...)` threading through `Propagator.create_initial_state`
- prompt blocks for Trader, the three risk debators, and Portfolio Manager (research-team prompts stay portfolio-blind); an explicit "not provided" notice when absent
- `--portfolio-context <json file>` CLI flag with fail-fast validation
- saved run state records `portfolio_context_present` plus the snapshot used
- checkpoint identity includes a deterministic portfolio-context fingerprint, so a changed snapshot starts a fresh run instead of resuming stale portfolio state
- renderer is market-neutral; optional currency is shown as an ISO-style label and quantities use generic units
- README documentation with a Python example

Note on open PR #1125 (read-only IBKR context for batch analysis): that change is broker-specific and batch-scoped, while this proposal is the broker-neutral counterpart on the canonical `propagate()` path. The two are complementary; happy to rebase or narrow scope per maintainer preference.

The design was informed by a downstream paper-trading integration where explicit missing-vs-empty portfolio semantics were required, but this PR keeps the upstream contract broker-neutral.

## Backward compatibility

Existing calls remain valid:

```python
ta.propagate("NVDA", "2026-01-15")
```

Portfolio context is optional. No new required env vars, no new runtime dependencies, no network calls in tests.

## Tests

- 55 new credential-free, network-free tests (`tests/test_portfolio_context.py`): schema validation, missing-vs-empty semantics, deterministic rendering, Trader / risk / Portfolio Manager prompt injection, `propagate()` wiring with backward-compat, saved-state metadata, CLI file loader, checkpoint fingerprint regressions, market-neutral rendering (USD / no-currency / JPY / crypto)
- Full suite: 708 passed, 2 skipped (baseline 653 passed before this change — no regressions)
- `ruff check .`, `compileall`, and `git diff --check` all green

### Checkpoint safety

Portfolio context participates in checkpoint identity through a deterministic hash.
A crashed run is resumed only when the portfolio snapshot is unchanged; if holdings or capital change, the next invocation starts a fresh run rather than continuing with stale state.

Related: #1166

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,37 @@ print(decision)
 
 See `tradingagents/default_config.py` for all configuration options.
 
+### Portfolio context (optional)
+
+`propagate()` accepts an optional, broker-neutral `PortfolioContext` snapshot describing current holdings and capital. When provided, the Trader, risk analysts, and Portfolio Manager ground their sizing language in it; when omitted, they are told explicitly that no portfolio context was given instead of assuming a flat portfolio. A context with empty `positions` means a known flat portfolio — this is distinct from providing no context at all.
+
+```python
+from tradingagents.agents.schemas import PortfolioContext
+from tradingagents.graph.trading_graph import TradingAgentsGraph
+from tradingagents.default_config import DEFAULT_CONFIG
+
+ta = TradingAgentsGraph(debug=True, config=DEFAULT_CONFIG.copy())
+
+context = PortfolioContext(
+    positions=[{"symbol": "NVDA", "quantity": 10.0, "average_entry_price": 150.0}],
+    cash=25000.0,
+    portfolio_value=120000.0,
+    as_of="2026-01-15",
+    source="paper-broker",
+)
+
+_, decision = ta.propagate("NVDA", "2026-01-15", portfolio_context=context)
+print(decision)
+```
+
+Existing calls without `portfolio_context` keep working unchanged. The CLI accepts the same snapshot as a JSON file:
+
+```bash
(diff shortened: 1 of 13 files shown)


Yijia-Xiao:
Thanks @ihsieh31. Implemented in 6436d1f: propagate() and the CLI take an optional portfolio, read by the trader, risk and portfolio agents; a run without one is never treated as a flat book.

PR #1324, opened 2026-09-11
fix(fred): keep the API key out of error messages and tracebacks

kbnnf:
## Problem

The FRED API key travels as a query parameter. `requests` builds its `HTTPError`
message from the full URL, so anything that catches and logs that error writes
the key out with it:

```
WARNING FRED series CP0000DEM086NEST unavailable
(403 Client Error: Forbidden for url:
 https://api.stlouisfed.org/fred/series/observations?series_id=...&api_key=<key in the clear>&file_type=json)
```

That was a real log line, produced by a caller doing nothing unusual — a
`try/except` around `_request` with a `logger.warning`. The same string reaches
an uncaught traceback, and from there CI output, a bug report, or a pasted
stack trace.

It is not a remote-exploitable bug: an attacker already able to read your logs
has other problems. It is a hygiene failure that makes a leak easy and silent,
and it fires most often exactly when someone is debugging and copying output
around.

## Fix

Add `redact()` and apply it on both error paths in `_request`:

- FRED's own `400` body, which can quote the request back;
- the `HTTPError` from `raise_for_status()`.

The re-raise keeps the `HTTPError` class **and** its `response` attribute, so
callers matching on the exception type or reading `exc.response.status_code`
for retry decisions are unaffected. `from None` avoids chaining the original,
unredacted exception into the traceback — without that, scrubbing the message
would accomplish nothing.

## Tests

Five cases in `FredKeyRedactionTests`, all mocked, no network and no key
required:

- `redact()` removes the key and leaves the rest of the URL readable
- text with no key passes through untouched
- a 5xx surfaces as an `HTTPError` whose message carries no key
- the re-raise keeps its class and its `response` (status code still readable)
- FRED's 400 body is scrubbed too

Full suite passes.

## Notes

Found while building a macro layer on top of this vendor, which fetches ~55
series per run — the failure showed up as soon as FRED's edge started
answering 403 on sustained bursts.

Unrelated but worth mentioning separately: FRED's documented limit is 120
requests/minute, but its WAF is stricter and returns a plain-HTML
`403 Access Denied` that outlives the minute and applies to the whole key
rather than the one request. Callers issuing series in parallel may want a
token bucket. Happy to open a second PR if that would be welcome.


Yijia-Xiao:
Thanks @kbnnf. Implemented in b20c8e6 for both key-authenticated vendors: API keys are removed from request errors.

PR #1329, opened 2026-09-12
fix(dataflows): honour omitted optionals in the Alpha Vantage global-news path

newnewself:
## Problem

`get_global_news` declares `look_back_days` / `limit` optional (JSON-schema default `null`) and forwards whatever it receives to the vendor. `yfinance_news` resolves `None` from config; the Alpha Vantage path fed it straight into `timedelta`:

```
get_global_news("2026-01-10", None, None)
-> TypeError: unsupported type for timedelta days component: NoneType
```

`news_data` is a core category (not in `OPTIONAL_CATEGORIES`), so `route_to_vendor` re-raises instead of degrading — the news analyst call aborts whenever Alpha Vantage serves global news and the model omits the optional args (which the tool description presents as the normal, config-inheriting usage).

## Fix

Resolve both optionals from `global_news_lookback_days` / `global_news_article_limit`, mirroring `get_global_news_yfinance`. Explicit values still win.

## Tests

`tests/test_alpha_vantage_hardening.py`: omitted args resolve from config (and produce the right window), explicit args still win.

Full suite: 656 passed, 1 skipped.


Yijia-Xiao:
Thanks @newnewself for the fix. Implemented in dffff22.

PR #1330, opened 2026-09-12
fix(dataflows): prune day-stamped OHLCV cache files

newnewself:
## Problem

The OHLCV cache filename embeds the request day:

```
{symbol}-YFin-data-{today - 5y}-{today + 1d}.csv
```

so every calendar day writes a new file per symbol. Nothing ever removes the old ones — the only cleanup in the repo is `clear_all_checkpoints` — so a long-lived install accumulates one file per symbol per day without bound.

## Fix

After a fresh cache file is written, drop the same symbol's files older than `OHLCV_CACHE_RETENTION_DAYS` (7). Only files matching that symbol's prefix are considered, and only those past the window, so a concurrent reader/writer or a racing prune is left alone; other symbols are untouched.

## Tests

New end-to-end test in `tests/test_ohlcv_cache_freshness.py`: an aged file for the same symbol is gone after `load_ohlcv`, while another symbol's aged file survives.

Full suite: 655 passed, 1 skipped.


Yijia-Xiao:
Thanks @newnewself. Fixed in 29e331a by keying the OHLCV cache by symbol, so it holds one file per symbol and needs no pruning.

Issue #1300, opened 2026-09-04
Look-ahead: get_fundamentals market snapshot stamped with wall-clock now(), not the analysis date


Rodvask:
## Summary

`get_fundamentals` in `tradingagents/dataflows/y_finance.py` stamps its header with `datetime.now()` (see lines ~274/331/366/401/436/465: `# Data retrieved on: {datetime.now()...}`). For symbols with no financial statements (e.g. crypto like `BTC-USD`), the entire fundamentals payload is just a market snapshot carrying **today's** date into a **historical** analysis — a point-in-time / look-ahead leak in the same family as #1275 (FRED), #1220 (social) and #1251 (memory), which v0.4.0 fixed elsewhere but not here.

## Evidence (reproduced on current main, post-v0.4.1)

Ran a point-in-time analysis for `BTC-USD` @ `2024-05-10` (crypto pipeline). The Fundamentals Analyst received a market snapshot stamped **2026-09-04** — over two years after the analysis date. Quote from the generated report:

> the fundamental-data vendor returned **no usable financial statements** for this symbol. The only usable data returned was a high-level market snapshot, and critically, **that snapshot is stamped with a retrieval date of 2026-09-04 — more than two years after our analysis date of 2024-05-10** — meaning the figures cannot be treated as the state of the market on the decision date.

(Credit: the analyst prompt correctly instructed the model to flag staleness instead of fabricating — the honest behavior worked, the data layer leaked.)

## Suggested fix

Thread the analysis date (`curr_date`) into the snapshot header the same way the OHLCV paths already pin their windows — i.e. stamp `# Data retrieved as of: {curr_date}` and, where possible, serve the snapshot fields (price, market cap, 52w high/low) from the as-of date instead of live quote endpoints. At minimum, a stale-dated header that says `:warning: snapshot is live, not point-in-time` would stop silent leakage into backtests.

## Environment

- main @ 9dee508 (post-v0.4.1 merge #1285)
- Symbol: BTC-USD, asset_type=crypto, trade_date=2024-05-10

Yijia-Xiao:
Thanks @Rodvask — the repro and the file:line trace made this quick to confirm. Shipped in `96111aa`.

[...]

PR #1331, opened 2026-09-12
fix(agents): require curr_date on the fundamental statement tools

newnewself:
## Problem

`get_balance_sheet`, `get_cashflow`, and `get_income_statement` declared `curr_date` optional, and the tool forwarded `None` when the model omitted it. The vendor-side look-ahead filter returns the frame unchanged for an unset date, so a historical run could receive fiscal periods dated after the analysis date. With a yfinance statement frame as of today:

```
curr_date omitted      -> ,2023-03-31,2024-09-30,2025-06-30   # future periods leak
curr_date=2023-06-30   -> ,2023-03-31
```

`get_fundamentals` already requires the argument; only the three statement tools let it be omitted.

## Relation to the merged profile fix (#1300)

Not the same thing. #1300 withholds the live company *profile* on a historical run, and its own docstring notes that "the statement tools stay point-in-time by filtering on `curr_date`". That guarantee only holds when the model actually passes the date — which the schema did not require. This PR makes that stated assumption real.

## Fix

Make `curr_date` required in the three tool schemas (moved ahead of the optional `freq`), and tell the fundamentals analyst to pass today's date. The vendor function signatures and their positional order (`ticker, freq, curr_date`) are unchanged.

## Behaviour note

This is an LLM-facing schema change: a call that omits `curr_date` is now rejected by the tool's argument validation so the model can retry, instead of being silently unfiltered. `tests/test_alpha_vantage_hardening.py::test_fundamentals_no_curr_date_passes_through` pins the vendor helper's own behaviour for a direct call without a date and is untouched.

## Tests

New `tests/test_fundamental_statement_dates.py`: the schema requires the date (`freq` stays optional), an omitted date is rejected, the tool forwards it to the vendor in the expected positional order, and the yfinance statement path drops post-date fiscal periods.

Full suite (rebased on v0.4.2): 683 passed, 1 skipped.


Yijia-Xiao:
Thanks @newnewself. Implemented in d04693a: every dated tool now takes the run's trade date from graph state, so an omitted or later date is bounded by it.

PR #1332, opened 2026-09-12
fix(cli): read and write the decision log on the CLI path

newnewself:
## Problem

The persistent decision log is only touched inside `propagate()` → `_run_graph()`. The CLI (`analyze` → `run_analysis`) builds its own initial state and streams `graph.graph` directly, so on a normal `tradingagents` run:

- no pending entry is written;
- prior entries are never settled, so no realised return / alpha / reflection is produced;
- the Portfolio Manager prompt never receives `past_context`.

README: *"The decision log is always on. Each completed run appends its decision to `~/.tradingagents/memory/trading_memory.md`. On the next run for the same ticker, TradingAgents fetches the realised return … and injects the most recent same-ticker decisions plus recent cross-ticker lessons into the Portfolio Manager prompt."* None of that happens on the primary entry point today.

This is the same shape as the `--checkpoint` bug fixed in #1249: lifecycle logic that lived only in `propagate()` was a no-op on the CLI. That fix introduced `begin_checkpoint` / `checkpoint_input` / `end_checkpoint` so both paths share the lifecycle; this PR does the same for the memory log.

## Fix

- `TradingAgentsGraph.prepare_memory_context(ticker, date)` — settles pending entries for the ticker and returns the date-gated past-context block (#1251 semantics preserved).
- `TradingAgentsGraph.record_decision(ticker, date, final_state)` — appends the decision; a run that produced no final decision (interrupted stream) is skipped with a warning instead of raising.
- `propagate()` now uses both; `run_analysis()` calls `prepare_memory_context` before building state and `record_decision` after a clean stream. A mid-stream failure records nothing, so the checkpoint is kept for resume and no partial decision is logged.

## Tests

New `tests/test_cli_decision_log.py`:

- the shared methods append a pending entry, skip an empty final state, and carry resolved lessons into the context block;
- a `run_analysis` run with a fake graph and patched UI asserts the initial state receives `past_context` and the completed run is recorded.

Full suite: 658 passed, 1 skipped.


Yijia-Xiao:
Thanks @newnewself. Implemented in 4a9f196: the CLI and propagate() share one decision-log path.

PR #1336, opened 2026-09-12
fix: handle explicit None for look_back_days/limit in alpha_vantage get_global_news

### LINKED ISSUE #1326: [Bug]: unexpected NoneType of timedelta
My code is on the latest version of be952b8, when I turn on news vendor of Alpha Vantage and run AMD from CLI interface, I got the following error:

> **Vendor 'alpha_vantage' failed for get_global_news: unsupported type for timedelta days component: NoneType.** 

It turns out that TradingAgents tool call passes **look_back_days=None** and **limits=None** to the function of **get_global_news(curr_date, look_back_days: int = 7, limit: int = 50)**, which disables the default values.

To fix the problem, I added a value check logic inside **get_global_news(...)** in **alpha_vantage_news.py ** as shown in the patch.

[alpha_vantage_news.patch](https://github.com/user-attachments/files/32137656/alpha_vantage_news.patch)

prithvirajrh:
## Summary
- The `get_global_news` LLM tool documents that passing `None` for `look_back_days`/`limit` inherits the configured default (see `tradingagents/agents/utils/news_data_tools.py`), but the Alpha Vantage vendor implementation only handled *omitted* arguments via literal defaults (`=7`, `=50`). When the LLM tool call explicitly passes `look_back_days=None`/`limit=None` (as reported in #1326), those literal defaults are bypassed and the call crashes inside `timedelta(days=None)`.
- This brings `alpha_vantage_news.get_global_news` in line with the existing convention already used by the yfinance vendor (`yfinance_news.get_global_news_yfinance`): resolve `None` via `get_config()`'s `global_news_lookback_days` / `global_news_article_limit`.

Fixes #1326.

## Test plan
- [x] Added `test_global_news_none_args_use_config_defaults` to `tests/test_alpha_vantage_hardening.py`, verifying `None` args resolve to config defaults and the vendor call succeeds instead of raising.
- [x] Confirmed the new test fails on `main` (pre-fix) and passes with the fix.
- [x] Full suite: `677 passed, 2 skipped` (skips are pre-existing, due to an optional dependency and a missing live API key, unrelated to this change).

https://claude.ai/code/session_01W8ioPGbswhsFA6EhBaZNbT


Yijia-Xiao:
Thanks @prithvirajrh for the fix. Implemented in dffff22.

Issue #1196, opened 2026-08-01
[Question] Debate layer shifts ambiguous/HOLD cases toward Overweight — does this reproduce elsewhere?


taro0915:
## Summary

  While testing locally (single local quantized model, trade date 2026-07-22), I ran the full graph (investment
  debate + risk debate, condition **C1**) against a single direct LLM call given the same underlying market/news
  report content with no debate structure (condition **C2**), and compared the distribution of final ratings.

  - **Experiment C** (AAPL, n=20 per condition, temperature at provider default): C1 was Overweight in 16/20 runs,
  C2 in 6/20 (80% vs. 30%, Fisher's exact p = 0.0036) — even though the market and news analyst reports feeding both
  conditions ended in `HOLD` in all 20 C1 runs.
  - **Experiment D** (AAPL, same date, n=10 per condition, `temperature=0.0` pinned for both): C1 was Overweight in
  6/10, C2 in **0/10** (all Hold). Fisher's exact p = 0.0108. So the separation survives removing sampling spread,
  and C2 became essentially deterministic while C1 did not.
  - **Experiment E** (TSLA — screened as clearly bearish, same date, n=10 per condition): **neither** condition
  produced a single bullish rating (C1 0/10, C2 0/10). Both sides landed on Underweight/Sell.

  Taken together these do not support "the debate layer is always bullish." The narrower description that fits all
  three is: the debate layer tips *ambiguous* cases — the ones where the analysts land on HOLD — toward the bullish
  side, and leaves a clearly bearish input bearish. See [Interpretation](#interpretation-updated-after-d-and-e) for
  what is observed vs. inferred.

  I separately filed #1176 (the first speaker in each debate round rebuts an opposing argument that has not been
  made yet, because the relevant "last response" state field starts empty). These come from the same run set and I
  mention it for context only — I have **not** run an experiment showing that #1176 is what produces the rating
  shift described here, and I am not claiming a causal link.

  This is 2 tickers, 1 date, 1 local quantized model, so I don't want to overstate it — posting as a
  discussion/question rather than a bug report: has anyone else measured something similar (or different) with other
  models, tickers, or dates?

  ## Experiment C — the original observation

  **Setup:** 20 iterations of the full graph including the Bull/Bear investment debate and the
  Aggressive/Conservative/Neutral risk debate (condition **C1**), and 20 iterations of a single direct LLM call
  given the same market/news report content and asked to output a rating with no debate structure (condition
[...]

PR #1337, opened 2026-09-13
fix(agents): stop conflict alone from defaulting the verdict to Hold (#1321)

### LINKED ISSUE #1321: [Bug] a4acd8a over-corrects #1196 — "materially conflicting" makes Hold the default
Long story short, Opus is convinced the verbiage change made more intelligent models more wishy-washy on rendering an actual verdict.  It was working mostly fine when I had the process working under a 35b model, but I recently switched to GLM 5.3 Flash and its become much more "conservative" in its decisions.

Many thanks for this amazing tool.  Keep it up please!
Obviously the below is AI written, but im terrible at communicating so...apologies

-------------------------------------------------------------------------------------------------------------------------------------------------------------

`a4acd8a` (#1196) added "materially conflicting" to the Hold condition in both manager prompts and both rating fields in `schemas.py`. But the pipeline *is* an adversarial debate — Bull vs Bear, Aggressive vs Conservative — so material conflict is present on every run, and the condition fires on every run.

My deployment went to **110 analyses with zero Buy/Overweight**, 86–89% Hold.

**Natural experiment.** I picked the change up in a version bump that went live on an image rebuild mid-day 2026-09-08. Same model (GLM-5.3-Flash, local vLLM) already running for two days, so only the prompt changed:

| 2026-09-08 | n | Hold |
|---|---|---|
| before rebuild | 6 | 0% |
| after rebuild | 23 | 70% |

Prior three weeks under the old wording: 0–35% Hold.

**Paired A/B.** Clause made runtime-switchable, same 13 ticker/date cells through both wordings, everything else identical:

| | Hold | coverage |
|---|---|---|
| `a4acd8a` | 11/13 | 15% |
| narrowed | 1/13 | 92% |

Sign test on discordant cells: 10–0, p=0.0020. (No profitability claim — 14 calls is far too few for that.)

**Suggested fix.** The intent of #1196 is right; just drop the term that swallows everything and say why:

> ...a bull/bear debate is adversarial by construction, so the mere presence of conflicting arguments is NOT grounds for Hold — resolving that conflict and saying which side wins is the job. Choose Hold only when the evidence is genuinely balanced after that weighing, or too thin to support any call.

Removing "materially conflicting" from all four sites alone would probably recover most of the range.

Not #1170 — structured output was well-formed throughout, all verdicts carried a matching `**Rating**:` header.

CadeYu:
## Summary

`a4acd8a` (#1196) added **"materially conflicting"** to the Hold condition in both manager prompts and both structured rating fields. The intent was correct — do not force a direction under ambiguity — but the pipeline *is* an adversarial debate by construction, so material conflict is present on every single run. The condition therefore fires every time, and Hold swallows directional calls.

One deployment measured **110 analyses with zero Buy/Overweight** and **86–89% Hold**. The issue reporter ran a paired A/B over 13 ticker/date cells: **11/13 Hold → 1/13 Hold** once the clause was narrowed (sign test *p* = 0.0020).

## What this changes

Four sites, all prompt/schema wording — no logic, no graph, no tests beyond the one added:

| File | What changed |
|---|---|
| `agents/managers/research_manager.py:39` | Prompt: drop "materially conflicting" as Hold trigger; add adversarial-by-construction framing |
| `agents/managers/portfolio_manager.py:65` | Same pattern for the risk-debate prompt |
| `agents/schemas.py` `ResearchPlan.recommendation` | Structured output description (models read these as instructions) |
| `agents/schemas.py` `PortfolioDecision.rating` | Same for the final portfolio manager rating |

**New principle:** an adversarial debate is the expected input. Conflicting arguments are not grounds for Hold — resolving that conflict and naming the winning side is the job. Hold is reserved for evidence that is *genuinely balanced after that weighing*, or too thin to support any call. The existing "do not manufacture a direction merely to appear decisive" clause is preserved unchanged.

## Regression test

`tests/test_hold_not_default_on_conflict.py` pins all four sites:

1. The old clause `materially conflicting` does not appear in either rendered prompt.
2. The phrase `adversarial by construction` IS present in both prompts.
3. Neither schema field description contains the old clause.
4. Both descriptions contain the explicit negation `not grounds for Hold`.

## Verified

```
680 passed, 2 skipped (bedrock not installed; deepseek key absent), 73 subtests — 2.97s
ruff check — all passed
```

Closes #1321.


Yijia-Xiao:
Thanks @CadeYu for the fix. Implemented in 62d3479.

PR #1342, opened 2026-09-13
fix: normalize 5-digit HK ticker codes to Yahoo's 4-digit form

### LINKED ISSUE #957: get wrong info with code 09992.HK
09992 is the code for popmart(泡泡玛特) in HK stock market, but in the report the team linked it with 农夫山泉,which is totally incorrect.

[complete_report.md](https://github.com/user-attachments/files/28558611/complete_report.md)

samhoooo:
## Summary
- Fixes #957: `09992.HK` (Pop Mart) was coming back for the wrong company (Nongfu Spring).
- Root cause: HKEX officially lists codes with up to 5 digits (`09992`), but Yahoo Finance only accepts the 4-digit form (`9992.HK`) — the 5-digit form 404s with no data. With every analyst tool returning empty, the agents fell back on their own training knowledge and hallucinated a company identity instead of surfacing the lookup failure.
- Fix: `normalize_symbol()` in `tradingagents/dataflows/symbol_utils.py` now strips leading zeros off `.HK` codes and re-pads to 4 digits before the request goes out, so `09992.HK → 9992.HK` and `00700.HK → 0700.HK`. Codes that are already 4 digits, like `0700.HK`, pass through unchanged.

## Verification
- Reproduced live: `yfinance.Ticker("09992.HK")` 404s ("Quote not found"); `yfinance.Ticker("9992.HK")` correctly returns Pop Mart International Group data.
- Added regression tests in `tests/test_symbol_utils.py` covering 5-digit repadding, 4-digit passthrough, and lowercase `.hk` suffix handling.

image

## Test plan
- [x] `pytest tests/test_symbol_utils.py -v` — 18/18 passed
- [x] `pytest -q` (full suite) — 679 passed, 2 skipped (unrelated, pre-existing)
- [x] `ruff check` on changed files — clean


Yijia-Xiao:
Thanks @samhoooo. Implemented in 99313bf: HK codes are zero-padded to Yahoo's 4-digit form.

PR #1347, opened 2026-09-13
fix(cli): persist decision memory after streamed runs

malandrindev:
## Summary

Fix CLI runs so completed streamed analyses persist the final trade decision to the decision-memory log before clearing the checkpoint.

## Problem

`TradingAgentsGraph.propagate()` persists the final decision via `memory_log.store_decision(...)`, but the CLI executes the graph directly with `graph.graph.stream(...)` and previously cleared the successful-run checkpoint without persisting the decision.

As a result, successful CLI runs could produce reports and checkpoints correctly while leaving `TRADINGAGENTS_MEMORY_LOG_PATH` untouched.

## Fix

- add `_finalize_streamed_run(...)` to merge streamed state deltas
- persist `final_trade_decision` through `graph.memory_log.store_decision(...)`
- clear the successful-run checkpoint only after memory persistence succeeds
- reuse the finalized state for the existing post-run reporting flow
- add a regression test that verifies both the persisted payload and the ordering: memory write before checkpoint cleanup

## Validation

- targeted regression suite: `128 passed`
- full non-integration suite: `677 passed, 1 skipped, 1 deselected, 73 subtests passed`
- Ruff: passed
- `git diff --check`: passed
- no-cost integration smoke with the real `TradingMemoryLog`: memory file created successfully and checkpoint cleanup invoked once

The change is intentionally limited to the CLI lifecycle and does not alter graph execution or provider behavior.


Yijia-Xiao:
Thanks @malandrindev. Implemented in 4a9f196: the CLI and propagate() share one decision-log path.

PR #1348, opened 2026-09-15
feat: Taiwan market foundation (TWSE/TPEx benchmarks, market profile, sentiment gating)

rita112025-cpu:
## Problems

Taiwan-listed tickers (`.TW` for TWSE, `.TWO` for TPEx) already price through yfinance, but the rest of the pipeline treats them like US stocks:

1. **Wrong alpha benchmark.** `benchmark_map` has no Taiwan entries, so the reflection layer measures `2330.TW` or `6488.TWO` against SPY instead of a Taiwan index.
2. **Misleading social sentiment.** The sentiment analyst always queries StockTwits and Reddit. Their coverage of Taiwan tickers is mostly empty or off-topic, and an empty feed can be read by the LLM as a neutral signal rather than as missing coverage.
3. **No market-level switch.** There is no single place to say "this ticker belongs to market X, which behaves differently", so any market-specific behaviour would end up as scattered suffix checks.

## Solution

Three small commits, each with tests. Every other market keeps its current behaviour.

1. **`.TW` benchmark** — add `".TW": "^TWII"` (TAIEX) to the existing `benchmark_map`. `_resolve_benchmark()` is unchanged; its exact `endswith` match already keeps `.T` (Tokyo) and `.TO` (Toronto) apart, and an explicit `benchmark_ticker` still overrides.
2. **Taiwan market profile** — add `market_profiles` to `DEFAULT_CONFIG` and a network-free `dataflows/market_profiles.py` (`resolve_market`, `get_market_profile`). The Taiwan profile covers `.TW` and `.TWO`, with per-suffix benchmarks (`.TWO` maps to `^TWOII`, the TPEx index), and a test keeps the profile consistent with `benchmark_map`.
3. **Sentiment gating by profile** — a profile can list `social_sources`; Taiwan sets `[]`. For a disabled feed the analyst sends no request and injects a "not enabled for this market ... not a neutral signal" placeholder, distinct from the existing "unavailable" and "no posts found" placeholders, and the prompt says missing coverage must lower confidence rather than imply neutral sentiment. Yahoo News is unaffected. Profiles without the key, and the default market, keep both feeds, and the analyst contains no market-specific branches.

## Testing

- `pytest`: 696 passed, 2 skipped (the existing optional `langchain_aws` and live DeepSeek skips).
- `ruff check .`: clean.
- The same diff passed this repository's CI workflow on my fork, on Python 3.10, 3.11, 3.12 and 3.13, plus ruff and the clean-install job: https://github.com/rita112025-cpu/TradingAgents/actions/runs/34936207560
- New tests cover `2330.TW` / `2454.TW` / `3443.TW` / `6488.TWO` / `5274.TWO` and lowercase input, non-Taiwan tickers (`7203.T`, `0700.HK`, `SHOP.TO`, `AAPL`) staying on their current benchmark and feeds, no StockTwits or Reddit request for Taiwan tickers, and the placeholder wording.

## Follow-ups (separate PRs, not included here)

- Official MOPS data for Taiwan tickers: monthly revenue and material announcements, point-in-time and fail-closed, together with the shared Taiwan helpers (clock, ticker-to-board parsing, as-of handling) they need.
- TWSE / TPEx institutional investor flows (三大法人) for the market analyst.


Yijia-Xiao:
Thanks @rita112025-cpu. Implemented in 47ac1f2: Taiwan tickers are measured against TAIEX.

PR #1352, opened 2026-09-15
feat(cli): add env-var overrides to skip report save/display prompts

### LINKED ISSUE #1133: feat: add env-var overrides to skip report save/display prompts
## Problem

When running TradingAgents in unattended mode (CI, cron, background scripts), the CLI blocks on two interactive prompts after analysis completes:
- "Save report?" 
- "Display full report on screen?"

There is no way to pre-answer these via environment variables, unlike other CLI steps which support `TRADINGAGENTS_*` overrides.

## Proposed Fix

Add two env vars following the existing `TRADINGAGENTS_*` convention:
- `TRADINGAGENTS_SAVE_REPORT=true` — skip save prompt, auto-save to default path
- `TRADINGAGENTS_DISPLAY_REPORT=true` — skip display prompt, auto-display

When unset, existing interactive prompts remain (no breaking change).

xiaodu55:
Closes #1133

## What

Adds two `TRADINGAGENTS_*` env overrides so unattended runs (CI, cron, background scripts) never block on the post-analysis prompts:

- `TRADINGAGENTS_SAVE_REPORT` — skips "Save report?" (and the save-path prompt); `true` saves straight to the default `reports/<ticker>_<timestamp>` path, `false` skips saving
- `TRADINGAGENTS_DISPLAY_REPORT` — skips "Display full report on screen?"

Both follow the existing env → `DEFAULT_CONFIG` overlay (`_ENV_OVERRIDES`), coerced with the same boolean rules as `TRADINGAGENTS_CHECKPOINT_ENABLED`. Unset variables keep the current interactive prompts (no breaking change).

## How

- `tradingagents/default_config.py`: two new `_ENV_OVERRIDES` rows plus `save_report` / `display_report` defaults (CLI-only keys; the engine never reads them)
- `cli/main.py`: the post-analysis save/display block moves into `maybe_save_and_display_report(final_state, ticker)` so it can be unit-tested; `run_analysis` calls it

## Tests

- `tests/test_env_overrides.py`: defaults + boolean coercion for the two new vars
- `tests/test_cli_env_skip.py` (`TestReportPromptsSkippedFromEnv`): env `true` auto-saves to the default path with zero prompts; `false` skips without saving; unset keeps all three interactive prompts; the display override works independently of the save override

Full suite: 683 passed. One pre-existing Windows-only failure in `test_api_key_env.py` (writes `.env` to `Path.home()` instead of cwd) is unrelated and reproducible on unmodified `main`.


Yijia-Xiao:
Thanks @xiaodu55. Implemented in 0c60284 as --save/--no-save and --show/--no-show flags rather than environment variables.

PR #1353, opened 2026-09-15
chore(deps): drop the unused backtrader and redis dependencies

shivsin25:
Both are declared as hard runtime dependencies in pyproject but are not imported anywhere in the repo:

    $ grep -rn "backtrader\|redis" --include="*.py" .
    (no matches)

backtrader pulls in a full event-driven backtesting engine and redis a client plus its parser stack, so every install — including the clean-install smoke job — pays to build and ship two trees nothing imports. Removing them narrows the dependency surface without changing behaviour; the full suite (676 passed, 2 skipped) is unaffected.


Yijia-Xiao:
Thanks @shivsin25. Done in ed6eae4: backtrader, redis and the runtime setuptools pin are gone.

Issue #1220, opened 2026-08-09
Historical analyses leak current Reddit and StockTwits data into the sentiment report


yiran1268:
### Summary

When `trade_date` is historical, the Sentiment Analyst correctly requests Yahoo news for the historical window, but still fetches the **current** StockTwits stream and Reddit's **current last week**. Those current posts are then placed in a prompt that says they cover the historical `start_date..end_date` window.

This introduces look-ahead bias and can make backtests use information that did not exist on the requested trading date.

### Reproduction

1. Run the sentiment analyst with a historical date, for example `trade_date="2020-01-15"`.
2. In `sentiment_analyst_node`, compare the three fetches:
   - Yahoo news receives `start_date="2020-01-08"` and `end_date="2020-01-15"`.
   - `fetch_stocktwits_messages(ticker, limit=30)` receives no date and calls the current symbol-stream endpoint.
   - `fetch_reddit_posts(ticker)` receives no date, and `_search_qs` hard-codes `t=week` (the current week).
3. The resulting prompt nevertheless says the report covers `2020-01-08` through `2020-01-15` and labels Reddit as “past 7 days”.

Relevant code:

- `tradingagents/agents/analysts/sentiment_analyst.py:62-71`
- `tradingagents/agents/analysts/sentiment_analyst.py:131-150`
- `tradingagents/dataflows/reddit.py:52-58`
- `tradingagents/dataflows/stocktwits.py:26,41-52`

### Expected behavior

Every observation used for an analysis should have a timestamp at or before `trade_date`. If the public social APIs cannot provide historical data, historical runs should omit those sources and inject an explicit “historical data unavailable” placeholder rather than substituting current posts.

### Actual behavior

Current social posts are included in historical analyses and presented to the LLM as though they came from the requested historical period.

### Suggested regression coverage

Add a test using an old `trade_date` that verifies neither social fetcher can return observations newer than `end_date` (or verifies that the sources are marked unavailable when historical retrieval is unsupported).

### Environment

- TradingAgents main at `01477f9` (v0.3.1)
- Python 3.12

yiran1268:
issue

Yijia-Xiao:
Thanks @yiran1268. Shipped in `9b98f09`: StockTwits and Reddit are now trimmed to the analysis window (and emit a clear placeholder when a historical window has none), so a backtest no longer shows today's chatter.

PR #1354, opened 2026-09-15
Feat/backtest harness

shivsin25:


shivsin25:
## What

Adds `tradingagents backtest`: replays the agent graph across a date range and
scores the resulting decisions as a portfolio, producing a markdown scorecard,
`equity.csv` and a JSON payload.

```bash
tradingagents backtest --tickers NVDA --start 2026-01-05 --end 2026-06-30
```

## Why

The last two releases were largely point-in-time correctness work, and each fix
was justified by a backtest:

| Fix | Rationale in the code |
| --- | --- |
| FRED vintage pinning (#1275) | *"would leak future information into a backtest"* |
| Social/news window (#1220) | *"rather than leaking today's chatter into a backtest"* |
| Memory PIT guard (#1251) | *"keeps a historical/backtest run from learning from outcomes that had not happened yet"* |
| stockstats cutoff | *"Filter to curr_date to prevent look-ahead bias in backtesting"* |
| Alpha Vantage fundamentals (#1300) | *"into a backtest"* |

But there was no way to run one. The CLI had a single interactive
ticker-and-date command, and nothing in the repo looped over dates. So that work
had no consumer, and there was no way to answer "does any of this help?"

This PR is the consumer. It doesn't reimplement point-in-time — it calls
`propagate()` in a loop and does the accounting, which is why it's mostly
additive.

## Design

**Producing decisions and scoring them are separate.** Running the agents is
slow and costs money; scoring is free. Each decision is appended to
`decisions.jsonl` as it is produced, and the scorer reads that back. Re-scoring
under different execution rules never re-runs an agent:

```bash
tradingagents backtest --tickers NVDA --start ... --end ... \
    --score-only --hold-policy flat --slippage-bps 15
```

The same property makes runs resumable — a later run with the same `--label`
skips stored decisions and retries failed ones. This sits above `--checkpoint`
(#1249), which recovers *within* a single decision.

The store is keyed by a signature covering models, analyst selection and debate
depth, so a config change starts a fresh experiment instead of blending two into
one equity curve.

**Fills are at the next bar's open**, never the decision bar's close. A decision
dated D is built from data through D, so filling at D's close would be the
look-ahead the rest of the framework prevents. `Hold` carries the position by
default, matching the Research Manager's definition of it. `REVIEW` (#1170) is
never traded and is counted separately.

**Baselines sit next to the headline number** — buy-and-hold of the benchmark,
of the traded names, and random re-orderings of the agent's *own* rating mix
through the identical execution model. Without that last one, an agent that
rates everything Buy in a rising market reads as skilled.

**`--strict-point-in-time`** installs a guard on the vendor dispatch path that
aborts any decision whose data requests reach past its trade date, turning the
look-ahead guarantees into an invariant that fails loudly. `route_to_vendor`
gained a `set_request_guard` hook for it; the hook is `None` by default and adds
no overhead on the normal path.

**Cost accounting.** Token usage is recorded per decision, and the scorecard
reports what a run cost alongside what it earned. No price table ships with the
framework — provider prices change and a stale table produces a confident number
nobody re-checks — so rates come from `config["llm_prices"]` or
`TRADINGAGENTS_LLM_PRICES`, and an unpriced model reports tokens without dollars.

## Changes outside the new package

- `StatsCallbackHandler` moved from `cli/stats_handler.py` to
  `tradingagents/stats.py`, so the package no longer imports from the CLI. The
  old import path still works.
- Benchmark resolution moved to `dataflows.symbol_utils.resolve_benchmark`,
  shared by the graph's deferred reflection and the scorecard so both measure
  alpha against the same index.
- `route_to_vendor` gained the optional `set_request_guard` hook.
- Adding a second CLI command would have made bare `tradingagents` fail with a
  missing-command error, since Typer treats a one-command app as *the* command.
  A callback keeps the original invocation running `analyze`, with a regression
  test.

## Testing

219 new tests, all offline — the graph, the price fetcher and the vendor dispatch
table are injected or stubbed, so nothing here needs an API key or network
access.

- `pytest -q` → **895 passed, 2 skipped** (676 on `main`)
- `ruff check .` → clean
- `pip install .` then `python -c "import tradingagents, cli.main"` → OK

## Scope / not included

Deliberately left out of a first harness: walk-forward optimisation,
cross-ticker risk budgeting, and a crypto trading calendar. Multi-ticker runs are
modelled as equal-capital sleeves with no rebalancing between them, which keeps
reported alpha attributable to the agent's per-ticker calls rather than to an
allocation scheme.

Happy to split this up or adjust any of the modelling defaults — particularly the
`Hold` policy and the weight map, which are judgement calls rather than facts.

## Note

Backtest output is research, not investment advice. The scorecard says so at the
top of every report, and the random baseline is prominent for the same reason: a
return figure with no reference point is not a result.

Tauric-Research:

PR #1356, opened 2026-09-16
fix(dataflows): filter global news before applying article limit

wxggzz:
## Summary

Apply the requested date window before counting global Yahoo news toward the article limit or deduplicating by title.

Previously, an initial search containing `limit` stale/future articles stopped subsequent queries before the date filter ran. The function could therefore return "No global news found" even when a later configured query supplied valid articles. An out-of-window headline could also suppress an in-window article with the same title.

## Changes

- Calculate the date window once before searching.
- Parse and date-filter each candidate before title deduplication and limit accounting.
- Continue configured queries until the requested number of unique, eligible articles is collected, or queries are exhausted.
- Reuse parsed article data when rendering the unchanged report format.
- If no eligible articles remain, return the explicit date-range no-news message, including when all searches are empty.

## Related work

This is a follow-up to #993 and #1008, not a new claim to the original discovery. The earlier fix added flat timestamp parsing and the explicit empty-report message, but the pre-filter limit/early-stop behavior remained on current `main` (`be952b8`).

## Validation

- Four new offline regressions cover flat and nested Yahoo payloads, later-query recovery after out-of-window candidates, deduplication order, and stopping at the unique eligible-article limit.
- All four new tests fail against the unmodified `main` implementation and pass with this change.
- `pytest -q`: **680 passed, 2 skipped, 73 subtests passed** on Python 3.12.13.
- The skips are the optional Bedrock dependency and a live DeepSeek API test; no live LLM or trading session was used to validate this patch.
- `ruff check .`: passed.
- `git diff --check`: passed.

## Scope

No vendor, retry policy, date-window semantics, or model prompt is changed. The Yahoo per-query `news_count` stays bounded by the requested limit. This fixes local filtering/selection order; it does not make Yahoo's live search a complete historical archive or eliminate all backtesting reproducibility limitations.

AI assistance: OpenAI Codex was used for investigation, implementation, testing, and review.


Yijia-Xiao:
Thanks @wxggzz for finding this. Fixed in 8d64416: each article is checked against the window as it is collected, so out-of-window items no longer spend the limit or stop the remaining searches.

PR #1357, opened 2026-09-17
fix(dataflows): render the Alpha Vantage news feed instead of forwarding it raw

### LINKED ISSUE #291: News data is extremely large / raw -> causing 6–7k token usage even with top-5 articles
Hi, I noticed that the news data block is extremely large. Even when selecting only the top 5 news articles, the input to the LLM ends up being 6–7k tokens, mainly because the full raw article text is passed through.

Other data sources (fundamentals, technicals) remain compact; only news is causing heavy token usage.

### **Problem**

* Very high token cost per run
* Slower inference
* News text contains many irrelevant sections (ads, disclaimers, long paragraphs)

### **Suggestion**

It might help to:

* Use only the **headline + first paragraph**, or
* Add a built-in summarization step, or
* Add a `max_chars`/`max_tokens` limit per article.

I’m currently using a workaround that keeps just the title + first paragraph, which reduces tokens significantly while keeping signal quality.

Thanks!

LudwigJMarx:
Closes #291.

## The problem

`_make_api_request` returns the response body verbatim, so the Alpha Vantage news vendor forwarded the raw `NEWS_SENTIMENT` document straight into the analyst's context. Every article in that document carries a banner-image URL, an author list, a topic array with relevance scores, `category_within_source`, `source_domain`, and **one sentiment entry per ticker it mentions** — a market-wide story can arrive with twenty.

Measured on a 20-article response with realistic summaries:

| | chars | ~tokens |
|---|---|---|
| raw payload (before) | 46,950 | ~11,700 |
| compact markdown (after) | 13,105 | ~3,300 |

**72% smaller**, with no field an analyst reads removed.

`get_news` also never sent a `limit`, so the endpoint served its default of 50 articles regardless of the configured `news_article_limit` — paid for, transferred, then discarded.

## The change

- Render the feed as the compact markdown the yfinance vendor already emits — `### {title} (source: {source}, {date})`, a sentiment line, the summary, the link — so both vendors hand the analyst the same shape.
- Keep only the requested ticker's sentiment entry plus the overall label; the other tickers' scores say nothing about the ticker under analysis.
- Pass the article limit to the endpoint instead of trimming after the transfer.
- Bound each article summary with a new `news_article_max_chars` (default 600, `0` disables). This was the issue's explicit ask, and it applies to **both** news vendors through a new `dataflows/news_format` so the rule doesn't fork per vendor — yfinance had no cap either.
- A body that isn't a feed document — a rate-limit notice, an unrecognised shape — is still returned verbatim rather than flattened into an empty report.

A sample of the new output:

```markdown
## NVDA News, from 2025-01-10 to 2025-01-16:

### Company beats on data-center revenue, raises outlook (source: Example Wire, 2025-01-15)
Sentiment: overall Somewhat-Bullish, NVDA Bullish (relevance 0.92)
Shares of the company moved after the latest quarterly report, which showed
data-center revenue climbing well past the guidance range management issued...
Link: https://www.example-wire.com/markets/story
```

## A crash on the same path

While covering this I hit a second, independent bug in the same function. The `get_global_news` tool documents both optional arguments as omittable ("omit to use the configured default") and forwards the omission as an explicit `None`, which overrode the literal defaults in the vendor signature:

```
TypeError: unsupported type for timedelta days component: NoneType
```

So with `news_data` set to `alpha_vantage`, any `get_global_news(curr_date)` call the agent made without the optional arguments crashed the run. Both now fall back to config the way `get_global_news_yfinance` already does. Happy to split this into its own PR if you'd rather keep the diffs separate.

## Testing

`tests/test_news_payload_size.py`, 11 cases: the noise fields are gone, the requested ticker's sentiment survives while the others don't, the output is under half the raw size, the limit reaches the endpoint *and* is enforced locally, empty and unrecognised bodies behave, the `None` arguments resolve, explicit arguments still win, the trim cuts on a word boundary and can be disabled, and the yfinance path trims too.

Full suite: 687 passed (was 676), `ruff check .` clean.


Yijia-Xiao:
Thanks @LudwigJMarx. The payload size is addressed in aef4af9 by requesting the configured article limit; we are not adding a separate rendering module and config key for the optional vendor.

PR #1358, opened 2026-09-17
fix(agents): always report a price target, and stop one bad price losing the decision

### LINKED ISSUE #1102: Price targets: how to force generate
Intuitively if there are 1 or 2 top signals I want to have would be **decision/rating and price target.**

But sometime price target is not included or generated in report. How do I force it to at best effort generate a range or some sort of price target? I would imagine I can edit the prompt or the code for the portfolio manager. Any ideas?

LudwigJMarx:
Closes #1102.

The issue asks how to force a price target to appear. Digging into it, there were two separate reasons it goes missing — and the second one is the more expensive of the two.

## 1. The line was dropped, not the value

`price_target` is optional and `render_pm_decision` omitted its line entirely when unset (same for `time_horizon`). So a reader could not tell *"no defensible target"* from *"this report never carries one"*, and anything consuming the decision had no stable section to read.

Both lines are now unconditional, with an explicit `Not provided` marker:

```markdown
**Rating**: Hold

**Executive Summary**: Hold the position; revisit after the print.

**Investment Thesis**: The bull and bear cases offset.

**Price Target**: Not provided

**Time Horizon**: Not provided
```

I deliberately **kept the schema field optional**, per @Ghraven's reasoning in the issue thread — making it required pushes a model into inventing precision on a weak-data run, which is worse than an honest blank. The fix is to make the *line* stable, not the value mandatory.

To actually raise the hit rate, the field description — which is what the model receives as its output instruction, per this module's own design note — now matches the standard its sibling `entry_price` already sets:

> Target price as an absolute number in the instrument's quote currency (e.g. 215.5), never a percentage and never a range — state the midpoint if you are reasoning about a band. Give your best estimate whenever the analysis supports one, even a rough level, and explain how you arrived at it in the investment thesis. Omit it only when the evidence cannot support any level, and say so in the thesis.

The prompt asks for it too, explicitly including the Hold case the issue reported as silently target-less.

## 2. One bad price was discarding the entire decision

This is the part I did not expect. `_coerce_optional_float` documents its own contract as leaving *"one bad field to null out instead of failing the whole proposal"* — but only placeholders and percentages were actually handled:

```python
>>> PortfolioDecision(rating="Hold", executive_summary="s", investment_thesis="t", price_target="$180-200")
ValidationError: price_target
```

A range, an approximation (`~215`), or a currency code (`USD 215`) raised — and since the validation error propagates out of `structured_llm.invoke()`, `invoke_structured_or_freetext` caught it and retried as free text. **One unreadable price threw away the rating, the executive summary and the thesis with it**, and the run silently lost its typed output. Nothing in the prompt told the model a range was forbidden, so it was being punished for a format it was never warned about.

Now:

| model wrote | before | after |
|---|---|---|
| `215.50`, `$1,234.50` | 215.5, 1234.5 | unchanged |
| `~215`, `about 215`, `USD 215`, `215 USD` | **raises** | 215.0 |
| `$180-200`, `180 to 200` | **raises** | `None` |
| `$215 by Q3 2026`, `high conviction` | **raises** | `None` |
| `N/A`, `15%` | `None` | unchanged |

Ranges are dropped rather than narrowed to one end, for the same reason a percentage is: they name two levels, and picking one invents a precision the model never stated — that is what the field description's "state the midpoint" is for. Anything still unreadable nulls that field alone.

The matching regex is anchored, so a number buried in prose does not match and the currency code must be genuinely uppercase (`USD 215` → 215, `the 215` → `None`) — guessing which number was meant is how a stop ends up on the wrong level.

The same helper backs the Trader's `entry_price` and `stop_loss`, so both agents gain the robustness.

## Testing

- `tests/test_pm_decision_report.py` (new, 8 cases): both lines render when unset and when set, every section present for every rating, a missing target is distinguishable from a present one, `0.0` is not mistaken for absent, the field description rules out percentages and ranges, the prompt asks for a level, and the line survives end-to-end into `final_trade_decision`.
- `tests/test_structured_agents.py` (+3 cases in the existing `TestNullishFloatCoercion`): qualified prices parse, ranges drop, and an unreadable price leaves the rating and thesis intact.

Verified `SignalProcessor` still extracts all five ratings from the new markdown, and the memory-log round-trip in `test_memory_log.py` is unaffected.

Full suite: 687 passed (was 676), `ruff check .` clean.


Yijia-Xiao:
Thanks @LudwigJMarx. Fixed in f8042ef: a price the model writes as a range or a hedge is dropped like any other unusable value instead of failing the whole decision, and an absent field is named as not provided.

PR #1361, opened 2026-09-17
test(config): lock in lazy dataflows config import (#194)

### LINKED ISSUE #194: Test suite hangs due to config loading before test environment setup
## Bug Description
The test suite hangs or times out when running unit tests due to a configuration loading issue.

## Root Cause
The `tradingagents/config.py` file calls `get_config()` at module import time (line 68), which validates API keys. However, the test fixture in `tests/conftest.py` sets test environment variables (lines 11-14) but imports DEFAULT_CONFIG afterwards (line 16). This creates a race condition where:

1. When pytest imports the test files, it loads conftest.py
2. conftest.py sets test environment variables
3. conftest.py then imports DEFAULT_CONFIG from tradingagents.default_config
4. This triggers the import of tradingagents.config, which immediately calls get_config()
5. The validation may fail or hang waiting for API keys

## Impact
- Unit tests cannot run properly
- CI/CD pipeline failures
- Development workflow blocked

## Solution
The config module should not eagerly load configuration at import time. Instead, it should use lazy loading or allow tests to override the configuration before validation occurs.

## Steps to Reproduce
1. Run `python -m pytest tests/unit/`
2. Tests will hang or timeout

## Environment
- Python 3.13
- pytest
- TradingAgents framework

Frank-zhu0404:
## Summary
#194 reported the suite hanging because config validated/loaded before test fixtures ran. On current `main` that hang is already gone (full suite ~676 passed without API keys; see discussion on the issue), and `tradingagents/dataflows/config.py` already keeps `_config = None` until first `get_config()` / `set_config()`.

This PR adds a regression test that reloads the module and asserts `_config is None` until first use, so import-time eager init cannot return unnoticed.

## Test plan
- [x] `pytest tests/test_dataflows_config.py`
- [x] Broader local suite previously green (~676 passed)

Closes #194


Tauric-Research:

Issue #557, opened 2026-04-17
[Data] Social Media Analyst reads Yahoo News, not social media — rename or fix


kelder66:
## Summary

The `social_media_analyst` is documented as analyzing "social media posts, recent company news, and public sentiment." However, its only tool is `get_news`, which routes to `yf.Ticker.get_news(count=20)` — the same Yahoo Finance news feed used by the News Analyst. There is no Reddit, StockTwits, X/Twitter, or any social media data consumed anywhere in the pipeline.

This creates two problems:
1. Users believe they are receiving social sentiment analysis when they are not.
2. The News Analyst and Social Media Analyst read from the same data source, making one of them redundant.

## Proposed solutions

**Option A — Rename:** Rename `social_media_analyst` to `sentiment_analyst` and update its prompt to focus on qualitative tone of recent news articles rather than claiming to analyze social platforms.

**Option B — Integrate real social data:** Add a new data tool backed by one or more of the following sources, all of which have free tiers:
- **Reddit** via PRAW (`r/wallstreetbets`, `r/investing`, `r/stocks`) — post volume, upvote ratio, sentiment per ticker
- **StockTwits** public API — message volume and bullish/bearish ratio (no auth required for read)
- **Google Trends** via `pytrends` — search volume proxy for retail interest

Example tool signature following the existing pattern:

```python
def get_social_sentiment(
    ticker: str,
    start_date: str,
    end_date: str,
) -> str:
    # returns post volume, sentiment ratio, top posts summary
    ...
```

## Why this matters

Retail sentiment on Reddit and StockTwits has demonstrated short-term price predictability in academic literature (Bollen et al., Ruan et al.). Currently the system does not consume this signal at all despite advertising that it does. Fixing the naming or the data would make the system more honest and more useful.

Yijia-Xiao:
Fixed in 0fcf136. Agent renamed to sentiment_analyst with real social data: pre-fetches news + StockTwits + Reddit (no API keys needed) and analyzes the actual data instead of fabricating it. Verified live on both OpenAI and Gemini — both now cite specific evidence (sentiment ratios, real quotes, upvote counts), zero fabrication patterns.

PR #1362, opened 2026-09-18
docs(uml): rename social_media_analyst node to sentiment_analyst

franklin-salcobian:
## Summary

Two Franklin-owned UML diagrams still named `social_media_analyst`, a module that no longer implements the agent. Renames the node to `sentiment_analyst` and regenerates the SVGs.

**Docs only — no source changes.** The investigation that prompted this concluded the opposite of what it set out to do; see below.

## Recommendation: keep the shim AND the deprecated alias

This started as a dead-code removal: delete `tradingagents/agents/analysts/social_media_analyst.py`, and decide whether `create_social_media_analyst` (`sentiment_analyst.py:198`) should go with it.

**The "unreferenced" finding is confirmed.** No importer of the module path exists anywhere:

- Repo-wide: only `sentiment_analyst.py`'s own docstring/alias and these two diagrams.
- Dynamic loading: every `importlib` / `__import__` / `sys.path` site in the repo is in `tests/`, reloading `default_config` or `cli.utils`. No `spec_from_file_location` anywhere. (Checked because DTA has a known sys.path-loading pattern invisible to grep.)
- Prod `/home/Franklin`: no reference outside the repo. No pip-installed copy in any `site-packages`.
- Cross-repo consumers — DTA `ai_agent_manager.py` and `integration/universe_dispatcher.py` — import only `TradingAgentsGraph` and `default_config`, never the analyst factories. So the alias has no external callers either.

**But neither is ours to delete.** Both files are byte-identical to the parent repo:

| File | our HEAD | `upstream/main` | `upstream/v0.4.0` |
|---|---|---|---|
| `social_media_analyst.py` | `8c72df0` | `8c72df0` | `8c72df0` |
| `sentiment_analyst.py` | `1bd11f2` | `1bd11f2` | `1bd11f2` |

The shim was authored upstream (`0fcf136`, 2026-05-11) and points at TauricResearch/TradingAgents#557. We are **0 commits behind `upstream/main`**.

A back-compat shim having no in-repo importer is not evidence that it is dead — that is its design; it exists for importers outside this tree. Since this fork pulls from the parent, deleting an upstream-tracked file buys 23 lines and costs a delete/modify conflict on every future upstream merge that touches it. Removing the alias would additionally diverge `sentiment_analyst.py`, a file we actively edit, making every upstream merge on it noisier.

**Let upstream retire its own deprecation; we inherit the removal on the next pull.** The real Franklin-owned drift was the diagrams, which is what this PR fixes.

## Second change: a diagram that never rendered

`07_package_diagram.puml` declares a **package** `"utils"` (line 32) and declared a **component** `[utils]` (line 90). PlantUML rejects the duplicate identifier, so `docs/svg/Package Diagram.svg` has been committed as a **7KB error dump instead of a diagram** — pre-existing, and reproducible from unmodified `origin/main`.

The cli component is now aliased (`component "utils" as cli_utils`, display label unchanged). The regenerated SVG is a real 96KB render. Without this the rename would have landed in the source but not in the published diagram.

Both SVGs regenerated with plantuml 1.2026.6.

## Verification

- `plantuml -tsvg` on both files: exit 0, no error output (previously: `Error line 90`).
- Regenerated SVGs contain `sentiment_analyst`, zero occurrences of `social_media`, zero error text.
- `git diff --name-only` touches only `docs/`.


franklin-salcobian:
Opened against the wrong repository by mistake — this is a Franklin-Digital fork-local docs change, not an upstream contribution. Reopening on the fork. Apologies for the noise.

PR #1366, opened 2026-09-18
feat: add brapi.dev vendor for B3 stock data and fundamentals

jccl2:
Description:
## Summary
- Adds the "brapi" vendor (brapi.dev) for OHLCV prices and fundamentals of B3 tickers (e.g., PETR4.SA), alongside yfinance and alpha_vantage.
- Registers the vendor in VENDOR_LIST and VENDOR_METHODS (get_stock_data, get_fundamentals), adds .SA -> ^BVSP to benchmark_map, and BRAPI_TOKEN to .env.example.
- Fixes validated against the real API: debtToEquity comes from financialData; the range bucket is chosen by the age from start_date to today, not by the start/end span.

## Known limitations
- get_verified_market_snapshot and get_indicators read yfinance directly (load_ohlcv) and do not go through the vendor router; brapi only powers get_stock_data and get_fundamentals.
- yfinance OHLC comes adjusted for dividends; brapi comes raw (only Adj Close is adjusted).
- dividendsData is fetched but not yet used in the fundamentals output.

## Test plan
- [x] py_compile and import smoke (brapi in VENDOR_LIST / VENDOR_METHODS)
- [x] 130 config/vendor/fundamentals tests
- [x] get_stock_data and get_fundamentals against the real API (PETR4.SA)
- [x] Complete TradingAgentsGraph with PETR4.SA on 2026-08-15 (Hold decision)


Yijia-Xiao:
Thanks @jccl2. The benchmark mapping is in 5c2dded: B3 tickers now measure alpha against the Ibovespa. We are not adding the vendor: yfinance already serves B3 prices and statements without a token.

PR #1368, opened 2026-09-18
test: isolate the suite from a developer's .env

codify88:
## What's wrong

`tradingagents/__init__.py` loads `.env` at **package import**:

```python
```

and `default_config.py` folds `TRADINGAGENTS_*` into `DEFAULT_CONFIG` at import
time, via `_apply_env_overrides`.

So for any contributor with a working `.env`, their own settings *are* the
framework defaults as far as the test suite is concerned. Two lines are enough to
turn the suite red:

```
TRADINGAGENTS_LLM_PROVIDER=anthropic
TRADINGAGENTS_QUICK_THINK_LLM=claude-sonnet-5
```

```
FAILED tests/test_cli_prefs.py::test_selections_are_remembered_after_a_run
  assert 'claude-sonnet-5' == 'gpt-5.6-mini'
FAILED tests/test_cli_prefs.py::test_a_remembered_endpoint_is_offered_back
  KeyError: 'default'
```

The second is the more interesting one: `TRADINGAGENTS_LLM_PROVIDER` makes
`get_user_selections` skip the provider prompt entirely, so the flow never reaches
`prompt_openai_compatible_url` and the test's recorder is never called. The
failure has nothing to do with the behaviour under test.

CI has no `.env`, so it never sees any of this.

## Why this is worth taking

**The failure lands on exactly the people you want contributing.** A `.env` is the
first thing anyone sets up — it is the documented way to supply keys. So the
contributors who hit this are the ones who got far enough to run the framework for
real, and what they see on their first `pytest` is a red suite. The natural
readings are "my change broke something" or "the repo is broken", and both cost
maintainer time in triage on an issue that cannot be reproduced in CI.

**It quietly weakens the tests it touches.** These assertions pin the *shipped*
defaults. Under a populated `.env` they are effectively asserting against whatever
that contributor configured, so a real regression in the defaults or in the prefs
round-trip would be masked for them and caught only in CI.

**It is contained.** Test-only, no production change, and no new dependency. The
overlay mechanism itself is untouched — `test_env_overrides.py` continues to
exercise it deliberately by setting its own variables and reloading the module,
and a clean slate is its intended starting point anyway.

## The fix

An autouse fixture strips `TRADINGAGENTS_*` from the environment for every test.

That alone is not enough, which is the subtle part: the overlay is applied at
*import*, so by the time any fixture runs, `DEFAULT_CONFIG` already holds the
developer's values and the shipped ones are gone. `conftest` therefore reloads
`default_config` once with the overlay stripped to recover the shipped values,
then puts the **pre-reload dict object** back as the module attribute — modules
that did `from tradingagents.default_config import DEFAULT_CONFIG` hold that
object, and rebinding the name alone would leave them on the overlaid copy. The
fixture then resets that dict per test and restores it afterwards.

## Verification

With a populated `.env` present:

| | before | after |
|---|---|---|
| `pytest -q` | 2 failed, 944 passed | **946 passed** |

Without `.env` (the CI condition): 946 passed, unchanged.

Reproduced on `2d17df8` (v0.5.0).

## Note

Filed alongside #1367, which fixes a separate, independent cause of local-only
failures in the OHLCV cache tests. On a machine that is both outside UTC and has a
`.env`, the two together account for all 10 failures; each PR fixes its own and
leaves the other's untouched.


Yijia-Xiao:
Thanks @codify88. Fixed in 1f7a2aa: every TRADINGAGENTS_* setting is blanked before the package loads, which also covers the values folded into DEFAULT_CONFIG at import.

PR #1371, opened 2026-09-19
fix(dataflows): scope the dataflows config per run instead of one process-wide dict

### LINKED ISSUE #1369: `set_config` merges into a process-wide dict, so a second graph inherits the first graph's `tool_vendors`
## Summary

`tradingagents/dataflows/config.py` keeps one module-level `_config`, and `set_config` merges dict-valued keys into it one level deep. `TradingAgentsGraph.__init__` calls `set_config(self.config)` (trading_graph.py:116). Two consequences for anyone who builds more than one graph in a process, which is what a vendor comparison or a backtest over configurations does:

1. A `tool_vendors` or `data_vendors` key set by one graph stays set for every later graph, including one built with the defaults. There is no way to unset it.
2. With two graphs alive at once (threads), both read whichever config was set last.

Data tools read this config at call time through `get_vendor`, so the run is served by vendors its own config never named, and nothing in the report says so.

## Repro (v0.5.0, 2d17df8, no network)

```python
import copy
from tradingagents.default_config import DEFAULT_CONFIG
from tradingagents.dataflows.config import set_config
from tradingagents.dataflows.interface import get_vendor

first = copy.deepcopy(DEFAULT_CONFIG)
first["tool_vendors"] = {"get_balance_sheet": "sec_edgar,yfinance"}
second = copy.deepcopy(DEFAULT_CONFIG)      # asks for the defaults

set_config(first)     # TradingAgentsGraph(config=first)
set_config(second)    # TradingAgentsGraph(config=second), same process
print(get_vendor("fundamental_data", "get_balance_sheet"))
# sec_edgar,yfinance      expected: yfinance
```

## How it was found

A matched pair of runs, same ticker and date, one with `get_fundamentals` routed to a new vendor and one without, executed two at a time in one process. The control run's tool log showed the treatment's vendor output. Both runs were discarded.

## Suggested direction

Smallest change: have `set_config` start from `DEFAULT_CONFIG` each time a full config is passed, so a graph gets what it asked for and nothing more. That fixes the sequential case and keeps the documented partial-update merge for callers who pass a partial dict on purpose.

Concurrent graphs need the config to travel with the run: a `contextvars.ContextVar` holding the graph's config, read first by `get_config`. LangGraph copies the context into its tool workers, so tool calls see the right one. A downstream app has run that way since the finding, with a regression test where two threads with different `tool_vendors` each read their own.

xiaodu55:
- Closes #1369

## What

`tradingagents/dataflows/config.py` kept one module-level `_config` and `set_config` merged dict-valued keys into it one level deep, so with more than one graph built in a process:

1. a `tool_vendors`/`data_vendors` key set by one graph stayed set for every later graph, including one built with the defaults — there was no way to unset it;
2. two graphs alive at once both read whichever config was set last, and data tools resolve vendors at call time through `get_vendor` → `get_config`, so a run could be served by vendors its own config never named.

Two changes, matching the direction suggested in the issue:

- `set_config` now rebuilds the active config from `DEFAULT_CONFIG` when the incoming dict carries every `DEFAULT_CONFIG` key (what `TradingAgentsGraph(config=...)` passes), so nothing leaks between graphs. Partial dicts keep the documented one-level merge.
- The resolved config is recorded in a `contextvars.ContextVar` that `get_config` reads first, so the config travels with the run (LangGraph copies the caller's context into its tool workers) and concurrent graphs in one process each read their own. The process-wide `_config` remains as the fallback for contexts that never set a config, and `tests/conftest.py`'s `_isolate_config` fixture now resets the context value alongside the global.

## How it was checked

- New tests in `tests/test_dataflows_config.py`: full-config rebuild drops a vendor override set by a previous graph; partial updates still merge onto the current config; two threads with different `tool_vendors` each read their own.
- Full suite: `pytest tests` → 942 passed, 88 subtests passed. The 4 failures (`test_api_key_env.py::test_ensure_api_key_prompts_and_writes_to_env`, 3× `test_ohlcv_cache_freshness.py`) also fail on unmodified `main` on this Windows machine (date/locale-sensitive), so they are pre-existing and untouched by this change.
- `ruff check` clean on the touched files.


davidalmeida90:
Tested this branch (4554021) against the repro in #1369: it now prints `yfinance`, as expected.

Also ran two threads that each call `set_config` with a different `tool_vendors` and then meet at a barrier before calling `get_vendor`: each read its own vendors (`yfinance` and `sec_edgar,yfinance`). `tests/test_dataflows_config.py`: 7 passed here, on Windows with Python 3.11.

Same approach, a `ContextVar` read first by `get_config`, has been running downstream since the finding with two graphs at a time, and LangGraph's tool workers do see the run's own config. Thanks for picking this up.

Yijia-Xiao:
Thanks @xiaodu55. Fixed in 96daaf1 (#1369). The config is bound when a run starts rather than when the graph is built, so a graph built before another still runs with its own.

PR #1373, opened 2026-09-20
ci: make the clean-install smoke test import from outside the checkout

boai-maker:
## The problem

`smoke-install` installs the package and then runs the import check **from the repository root**:

```yaml
pip install .
python -c "import tradingagents, cli.main; print('clean-install import OK')"
```

`python -c` prepends the current working directory to `sys.path`. At the repo root that directory contains `tradingagents/` and `cli/`, so both names resolve to the **source tree**, not to the distribution pip just installed.

Verified on a clean 3.12 venv against current `main`:

```
# from the repo root — what CI does today
cli           -> <repo>/cli/__init__.py
tradingagents -> <repo>/tradingagents/__init__.py

# from outside the checkout
cli           -> site-packages/cli/__init__.py
tradingagents -> site-packages/tradingagents/__init__.py
```

The package is packaged correctly today, so **nothing is currently red** — this is a latent gap, not a live break. But the job cannot fail for the reason it exists. A module dropped from the built artifact, or an undeclared runtime dependency (the `#994` python-dotenv case the comment names), would still be satisfied by the source tree in the working directory and the job would report success.

`cd "$RUNNER_TEMP"` before the import makes it test the installed distribution.

## Also hardened while touching the file

| Change | Why |
|---|---|
| `permissions: contents: read` | No job writes to the repo, packages, or checks — don't inherit the repository default |
| `concurrency` keyed on workflow name, `cancel-in-progress` only for `pull_request` | Every commit landing on `main` keeps a CI result instead of being cancelled by the next push |
| pip caching via `setup-python`, keyed on `pyproject.toml` | The four matrix legs currently rebuild the dev env from scratch each run |
| `ruff==0.16.8` instead of `ruff>=0.15` | That constraint already floats a full minor — it resolves to **0.16.8** today. `ruff check .` still passes on it, so this isn't a fix for a break; it stops a future release adding a rule to the strict select and reddening `main` with no commit behind it |
| `timeout-minutes` on all three jobs | A hung `pytest` is cut off in 20 min rather than consuming the 6-hour default |

## Deliberately not included

`ruff format --check` — `pyproject.toml` notes that whole-repo formatter adoption is deferred until the open-PR backlog clears, and it would reformat 118 of 160 files today. Left alone.

No change to the rule set, the matrix, or what is linted.

## Verification

- `ruff check .` passes on 0.16.8 (unchanged behaviour)
- Workflow parses as valid YAML; all three jobs intact
- Import shadowing reproduced and the fix confirmed, paths above


boai-maker:
Closing this — withdrawing at the author's request. Apologies for the noise.

